PR #8, opened 2017-02-14
Adding macOS development & test instructions

liyanchang:
Two documentation commits:

1. Add macOS instructions. Not terribly different from the Ubuntu instructions but figured it was worth calling out. I tried to format it similarly to the README and how it dealt with multiple OSes. Feel free to let me know if you'd prefer another format or just patch it.

2. A line on how to run tests. Most folks should know how to, just by looking but why not call it out. If you spent the time writing tests, might as well have people run them.

Thanks!

### DIFF
--- a/doc/development.adoc
+++ b/doc/development.adoc
@@ -1,27 +1,45 @@
-== Setting up a development environment on Ubuntu 16.04 (Xenial)
+= Setting up a development environment
 
-Install development dependencies:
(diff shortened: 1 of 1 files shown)


dagheyman:
Thanks. I've decided not to merge this right now as it adds more noise and I don't want to maintain docs for another development environment right now. The line about running the tests probably make sense though, might add it later.

PR #10, opened 2017-02-25
Strip out spaces from OATH base32 code entry

liyanchang:
Some services (e.g. Dropbox, Google) will have spaces in the base32
code. They do it because base32 does not contain spaces and it makes it
easier for a human to read and type.

In the case of ykman, it's likely that folks copy and paste the code and
thus will also copy over the spaces. As it is safe to ignore the spaces,
we should.

### DIFF
--- a/test/test_cli_commands_on_yubikey.py
+++ b/test/test_cli_commands_on_yubikey.py
@@ -177,6 +177,11 @@ class TestOATH(unittest.TestCase):
         creds = ykman_cli('oath', 'list')
         self.assertIn('test-name', creds)
 
+    def test_oath_add_credential_with_space(self):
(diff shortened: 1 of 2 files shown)


dagheyman:
Thank you, I agree with you that whitespace inside the secret key should be stripped. 

I'd suggest we do it inside the `parse_b32_key` function instead, where we already do some input cleaning (we convert all input to uppercase). That way it's not only the prompt entry that gets this fix, but all base32 key input in the application. If you update this PR with that change I'll be happy to merge this.

liyanchang:
Thanks! Also for the note on the precommit hook. Did not install that and took an extra minute to figure out why I broke the build.

PR #22, opened 2017-09-04
add repl support

### LINKED ISSUE #21: REPL mode
Currently using this on a low-resource system (Raspberry Pi), where each ykman invocation takes double-digit seconds to run, as it loads up all the relevant libraries and such. Would rock if I could launch a REPL and run the commands in sequence instead of paying the load price w/ each invocation

akerl:
This fixes #21 . It does expect a yubikey that has all modes enabled, since it can't do cuteness in the transport checks given it doesn't know what commands the user will run, but I'm not sure how to set up that kind of check w/o more smarts

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ from ykman.yubicommon.setup import setup
 
 
 install_requires = [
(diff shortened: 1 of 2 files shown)


akerl:
hm, it seems like this doesn't play nice with the USB handling logic :(

PR #25, opened 2017-10-02
Updated macOS Install instructions

rickard-von-essen:
Added ` brew install ykman` as the main installation method.

Also changed to use _Python 2_  as the default instructions for installing from source since that is (still) the default on macOS.


emlun:
Looks good! Thanks!

PR #36, opened 2018-01-10
OpenPGP: print remaining PIN retries in info cmd

dagheyman:
CLI: Print remaining PIN retries in `$ ykman openpgp info` command
API: make `get_remaining_pin_tries()` a public method.

### DIFF
--- a/ykman/cli/opgp.py
+++ b/ykman/cli/opgp.py
@@ -98,9 +98,9 @@ def info(ctx):
     controller = ctx.obj['controller']
     click.echo('OpenPGP version: %d.%d.%d' % controller.version)
     retries = controller.get_remaining_pin_tries()
-    click.echo('PIN tries remaining: {}'.format(retries['pin']))
-    click.echo('Reset code tries remaining: {}'.format(retries['reset']))
-    click.echo('Admin PIN tries remaining: {}'.format(retries['admin']))
+    click.echo('PIN tries remaining: {}'.format(retries.pin))
+    click.echo('Reset code tries remaining: {}'.format(retries.reset))
+    click.echo('Admin PIN tries remaining: {}'.format(retries.admin))
 
 
 @openpgp.command()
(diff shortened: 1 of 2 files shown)


ykman/cli/opgp.py
```diff
@@ -97,6 +97,10 @@ def info(ctx):
     """
     controller = ctx.obj['controller']
     click.echo('OpenPGP version: %d.%d.%d' % controller.version)
+    retries = controller.get_remaining_pin_tries()
+    click.echo('PIN tries remaining: {}'.format(retries[0]))
+    click.echo('Reset code tries remaining: {}'.format(retries[1]))
+    click.echo('Admin PIN tries remaining: {}'.format(retries[2]))
```
emlun: I feel like we should name these values instead of addressing them by index, where possible. How about renaming `get_remaining_pin_tries()` to `_get_remaining_pin_tries()`, and adding a new `get_remaining_pin_tries()` which returns a dict instead? (I can push a prototype of that)

emlun: See 4f77e92d8225365538e861b7100c21f214713ff8

dagheyman: Agreed.

dainnilsson: How about a namedtuple?

emlun: @dainnilsson Even better!

dainnilsson:
Is there value in still having _get_remaining_pin_retries() instead of just the single public method? If not I'd just drop it.

emlun:
I'll see if I can eliminate the index parameter from https://github.com/Yubico/yubikey-manager/blob/4f77e92d8225365538e861b7100c21f214713ff8/ykman/opgp.py#L143

dainnilsson:
AFAIK a namedtuple `is a` tuple, so indexing should still be possible.

emlun:
Oh right, of course. I was apparently still half thinking about dicts.

emlun:
@dagheyman Merge when ready!

PR #38, opened 2018-01-18
CCID: only try readers with yubico in it

dagheyman:
This seemed to been have dropped by mistake in a refactor.


dagheyman:
Dropped in https://github.com/Yubico/yubikey-manager/commit/7b329469ea45196fb262812cab70af5f57ed4b33#diff-eac76d44800ad2fb0ea99a710ce4ff0d

emlun:
Or perhaps we could filter the list in `_list_readers()` instead?

PR #39, opened 2018-01-22
Fixes for programming OTPs in the prompt

dagheyman:
Providing the public id, private id and secret key from the interactive prompt was broken.


dagheyman:
There seems to be [a way](http://click.pocoo.org/5/testing/#input-streams) for testing the prompts, but I'll add that in a separate PR.

PR #41, opened 2018-01-22
Log lightly handled errors

emlun:
I went through all `except` statements and added error logging where it seemed appropriate. My basic rules of thumb for when this was not needed were:

- If the `except` block (or the code following a `pass` statement) attempts to recover from the error
- In `except` block branches that raise new errors
- If the `except` block is on the CLI level and appropriate feedback is already given

### DIFF
--- a/ykman/cli/oath.py
+++ b/ykman/cli/oath.py
@@ -27,6 +27,7 @@
 
 from __future__ import absolute_import
 import click
+import logging
 from threading import Timer
 from binascii import b2a_hex, a2b_hex
 from .util import (
@@ -37,6 +38,9 @@ from ..driver_ccid import APDUError,  SW_APPLICATION_NOT_FOUND
 from ..util import TRANSPORT, parse_b32_key
 from ..oath import OathController, SW, CredentialData, OATH_TYPE, ALGO
 from ..settings import Settings
+
+
+logger = logging.getLogger(__name__)
 
 click_touch_option = click.option(
     '-t', '--touch', is_flag=True,
@@ -474,7 +478,8 @@ def ensure_validated(ctx, prompt='Enter your password', remember=False):
             try:
                 controller.validate(a2b_hex(keys[controller.id]))
                 return
-            except Exception:
+            except Exception as e:
+                logger.debug('Error', exc_info=e)
                 del keys[controller.id]
 
         # Prompt for password
--- a/ykman/cli/slot.py
+++ b/ykman/cli/slot.py
@@ -35,9 +35,13 @@ from ..util import (
     modhex_encode, parse_key, parse_b32_key)
 from binascii import a2b_hex, b2a_hex
 from ..driver_otp import YkpersError
+import logging
 import os
 import struct
 import click
+
+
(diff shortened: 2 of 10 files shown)


ykman/cli/oath.py
```diff
@@ -474,7 +478,8 @@ def ensure_validated(ctx, prompt='Enter your password', remember=False):
             try:
                 controller.validate(a2b_hex(keys[controller.id]))
                 return
-            except Exception:
+            except Exception as e:
+                logger.debug('Error', exc_info=e)
                 del keys[controller.id]
```
emlun: What does line 483 actually do in practice?

ykman/cli/slot.py
```diff
@@ -54,7 +58,8 @@ def inner(ctx, param, val):
                                      callback=lambda c, p, v: int(v))
 
 
-def _failed_to_write_msg(ctx):
+def _failed_to_write_msg(ctx, exc_info):
+    logger.error('Failed to write message to driver', exc_info=exc_info)
```
dagheyman: It didn't fail to write a message, the function name is referring to that it failed to write and should give the user a message. So I think the log error message should be something like "Failed to write to the device".

ykman/driver_ccid.py
```diff
@@ -102,8 +102,9 @@ def __init__(self, connection, name):
         self._pid = _pid_from_name(name)
         try:
             self._read_serial()
-        except APDUError:
-            pass  # Can't read serial or version
+        except APDUError as e:
+            # Can't read serial or version
+            logger.error('Failed to read serial number', exc_info=e)
```
dagheyman: Failed to read serial or version? Or is the comment outdated?

ykman/piv.py
```diff
@@ -588,9 +592,8 @@ def set_mgm_key(self, new_key, touch=False, store_on_device=False):
                 self.put_data(
                     OBJ.PIVMAN_PROTECTED_DATA,
                     self._pivman_protected_data.get_bytes())
-            except APDUError:
-                # No pin provided, can't clear key..
-                pass
+            except APDUError as e:
+                logger.debug("No pin provided, can't clear key..", exc_info=e)
```
dagheyman: pin -> PIN

emlun:
I agree about that being a bit redundant. The reason I added those log statements was so there's some way to get the exception backtrace out, since the plain echo functions don't include that and the exception isn't re-thrown to be printed somewhere else.

ykman/cli/slot.py
emlun: Ah, good catch!

ykman/driver_ccid.py
emlun: I wasn't sure, judging by the method name, so I left the comment in place. But it turns out the `_read_serial` method does indeed read both serial number and version, so I'll refactor that.

ykman/piv.py
emlun: 👍

dagheyman:
Any feedback on this one @dainnilsson ?

PR #44, opened 2018-01-29
Fixes for opening device on slow systems

dagheyman:


ykman/descriptor.py
```diff
@@ -137,7 +137,7 @@ def list_drivers(transports=sum(TRANSPORT)):
                 yield dev
 
 
-def open_driver(transports=sum(TRANSPORT), serial=None, pid=None, attempts=3):
```
dainnilsson: Can we make sleep_time here grow exponentially instead of linearly? With that the number of attempts can probably be lowered.

PR #46, opened 2018-01-31
Logging improvements

emlun:
- Disable logging by default
- Make logging_setup module usable by dependent projects, instead of making them duplicate it


emlun:
The stripping is needed because the QML adapter [serializes the arguments to JSON][ykqml], which encloses strings in quotes. Python's `str.strip` replaces all (or a maximum number counting from the start) occurences of the targeted string, but what we want here is instead to remove quotes only from the start and end and only if they both exist.

[ykqml]: https://github.com/Yubico/yubioath-desktop/blob/master/qml/YubiKey.qml#L70

emlun:
"Accept any-case log level name" currently doesn't carry through to the CLI because the `type=click.Choice(...)` still only lists the uppercase options, but at least the support is there now. It should be exposed to the ykman-gui and yubioath-desktop CLIs (which don't validate the option argument), for instance.

dagheyman:
Happy to merge this if @dainnilsson gives thumbs up.

dainnilsson:
On the quite-stripping: The QML code does serialize arguments as JSON, but typically the inverse is done on the python side. Isn't this the case here? If not, why?

emlun:
Good question... It looks like both serialization and deserialization happens in the [Controller][controller] in `yubikey.py` in the dependent project. I suppose we could call this code via the `Controller`, too, but then there would be a race condition between initializing the logging framework and calling log methods in other Python code (logging early in the `Controller` might not appear, for example).

I guess we could have two separate init functions in `logging_setup.py`, one accepting raw Python arguments and one accepting JSON. It would seem a bit overkill to do that for just the `log_level` parameter, but perhaps it could be worthwhile for the `log_file_name` parameter?

[controller]: https://github.com/Yubico/yubioath-desktop/blob/master/py/yubikey.py#L91

dainnilsson:
There's also [this](https://github.com/Yubico/yubikey-manager-qt/blob/master/ykman-gui/py/logging_setup.py) which looks to be replicating things.
As a general comment I'd say that as the JSON serialization is an implementation detail in the QML code, the deserialization (which should probably use json.loads for consistency) should be done in the same project to not contaminate this one.

emlun:
Yes, that file will be deleted as a step in upgrading to this.

I'm inclined to agree about the contamination thing, I'll make minimal (de)serialization wrappers in the dependent projects instead.

PR #48, opened 2018-02-02
Set Debian version to 0.6.0~dev

emlun:
To agree with version in `__init__.py`.


dagheyman:
Doesn't look like travis is very happy about this.

emlun:
Oops...

PR #49, opened 2018-02-05
Drop yubicommon submodule

dagheyman:
What:
 - Drop dependecy on yubicommon
 - Move in ctypes code to native/modifiedctypes
 - Explicitly list packages in setup.py

Why:
 - Simplify package structure
 - No hacks to delete symlink needed anymore on Windows
 - Less magic in setup.py


dainnilsson:
Don't really like the "modifiedctypes" name. It sounds like it's not actually ctypes, but a modification of it. Can't these files just go in the ykman.native package? Just rename the __init__.py to something else, so you have ykman.native.libloader and ykman.native.ctypesutils or something.

dagheyman:
> Don't really like the "modifiedctypes" name. It sounds like it's not actually ctypes, but a modification of it. Can't these files just go in the ykman.native package? Just rename the init.py to something else, so you have ykman.native.libloader and ykman.native.ctypesutils or something.

Sure, I'll update the PR.

PR #53, opened 2018-02-08
Require explicit YubiKey serial number for destructive tests

emlun:
This replaces the `INTEGRATION_TESTS=TRUE` environment variable with `DESTRUCTIVE_TEST_YUBIKEY_SERIAL=<serial number>`. It also adds the environment variable `DESTRUCTIVE_TEST_DO_NOT_PROMPT=TRUE` which skips the prompt before proceeding to destroy the key.

### DIFF
--- a/test/test_cli_commands_on_yubikey.py
+++ b/test/test_cli_commands_on_yubikey.py
@@ -1,5 +1,4 @@
 import os
-import sys
 import unittest
 import time
 import click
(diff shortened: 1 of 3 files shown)


dagheyman:
Initial testing: when i put a serial number that is wrong in the variable (like just 1), all the tests try to run anyway and the tests get kind of stuck. Maybe it should do an initial check that that device is present?

emlun:
Ah, good catch! I did that kind of filtering at first, but took it out because it slows startup down quite a bit even when the key is present. But evidently I forgot to test this case, so I can put it back in.

And apparently I _also_ lost the `== 1` when setting `_one_yubikey`...

emlun:
@dainnilsson remarks in #54 that

>it would be non-obvious that the descriptor opens a connection to the device

which is a concern that might be applicable to the [get_descriptors_with_serials()][gdws] here as well. What do you think?

[gdws]: https://github.com/Yubico/yubikey-manager/pull/53/files#diff-1758d235dba6f37ac90b71e9ae360bbbR129

dainnilsson:
It could be, but less so I think. More of a concern is that all the descriptors returned are created using .from_driver, which in some cases will lose information about the actual firmware version (it's always reliably readable from the USB descriptor, but not from the device).

emlun:
It turned out that `get_descriptors_with_serials()` is unnecessary, so I reverted its addition. This PR now only changes `test_cli_commands_on_yubikey.py`.

dagheyman:
One more small thing: testing with a serial number that is not present currently spits out FailedOpeningDeviceException + full stack trace, maybe it can give an nicer error message.

emlun:
@dagheyman How about now? :)

dagheyman:
Nice. Any reason the import statements are not at the top of the file?

emlun:
Not really, the one in the `if` statement was there when I started editing this.

PR #54, opened 2018-02-08
Attempt to wink YubiKey when prompting for destructive test confirmation

emlun:
For some extra visual feedback.

Note that this PR is relative to #53, not master.


dainnilsson:
I'm not really a fan of this in general, it seems like it's more of a "because it's cool" thing than useful, imo.

If done, I don't like it as a method on descriptor, as it would be non-obvious that the descriptor opens a connection to the device, which can interrupt other things. It's only available for the U2F driver, so it should only be a exposed as a method there.

emlun:
I can understand those sentiments, and partly share them. I'm not against dropping this PR if you prefer keeping it out.

dagheyman:
I agree, I don't think this use case justifies adding wink to the API.

PR #57, opened 2018-02-15
Add venv setup instructions to development.adoc

emlun:


dagheyman:
Any specific reason? I'm thinking if you develop in python and enjoy using venvs, you find this information elsewhere. I think our docs should describe stuff specific for this project and be as short as possible.

emlun:
Good point. Do we close this, then?

PR #58, opened 2018-02-15
Clean up tests

emlun:
The integration tests now reliably pass in both Linux and Mac!

### DIFF
--- a/test/test_external_libs.py
+++ b/test/test_external_libs.py
@@ -1,10 +1,7 @@
 import unittest
-import os
 from test.util import ykman_cli
 
 
-@unittest.skipIf(
-    os.environ.get('INTEGRATION_TESTS') != 'TRUE', 'INTEGRATION_TESTS != TRUE')
 class TestExternalLibraries(unittest.TestCase):
 
     def test_ykman_version(self):
--- a/test/test_external_libs.py
+++ b/test/test_external_libs.py
@@ -1,7 +1,10 @@
 import unittest
+import os
 from test.util import ykman_cli
 
 
+@unittest.skipIf(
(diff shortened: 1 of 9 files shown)


ykman/device.py
```diff
@@ -180,6 +184,10 @@ def use_transport(self, transport):
         assert dev.mode == my_mode
         return dev
 
+    def close(self):
+        logger.debug('close')
```
dagheyman: Remove this log message or improve it a bit.

ykman/driver.py
```diff
@@ -64,3 +68,6 @@ def read_capabilities(self):
     def guess_version(self):
         # Second arg is True if the version is certain, False if not.
         return (0, 0, 0), False
+
+    def close(self):
+        logger.debug('close')
```
dagheyman: See above (about log msg)

test/test_external_libs.py
```diff
@@ -1,10 +1,7 @@
 import unittest
-import os
 from test.util import ykman_cli
 
 
-@unittest.skipIf(
-    os.environ.get('INTEGRATION_TESTS') != 'TRUE', 'INTEGRATION_TESTS != TRUE')
```
dagheyman: Are you sure about this? Shouldn't the python only unittests be able to run without the libraries in place?

dagheyman:
Very nice, I look forward to approving :)

test/test_external_libs.py
emlun: Oooh, now I see why this was there. Good catch, thanks! This should just be updated to use the new environment variable, then.

emlun: Case in point: all CI build failed... 😁

PR #59, opened 2018-02-16
Use a VERSION file as source of truth for version

dagheyman:
I went for [option number 4](https://packaging.python.org/guides/single-sourcing-package-version/), mostly because we can then reuse the VERSION file thing for other non python projects. And less regex.


dainnilsson:
I believe the VERSION file will have to be in the ykman package for this to actually work. I don't think it gets installed as part of the package otherwise.

dagheyman:
> I believe the VERSION file will have to be in the ykman package for this to actually work. I don't think it gets installed as part of the package otherwise.

I only tested with `python setup.py sdist` and `pip install -e` apparently. Now the version file lives inside the ykman folder, and seems to be included on install thanks to `package_data`. `MANIFEST.in` is still needed to include it in `sdist`.

I don't like this as much as a version file in the root directory. Oh well.

PR #60, opened 2018-02-20
Split test_cli_commands_on_yubikey into multiple modules

emlun:


dagheyman:
I think it might be a good idea to name all test files according to the pattern `test*.py`. See https://docs.python.org/3/library/unittest.html#cmdoption-unittest-discover-p . Without this `python -m unittest` doesn't find them.

dainnilsson:
@dagheyman Good point!

emlun:
Good catch, thanks!

PR #62, opened 2018-02-20
Log version after initializing logging

emlun:
Mostly for de-confusion during development, but could possibly be useful for bug reports too. This carries through to the GUIs so this will print the ykman library version if it happens to be different from what you'd expect from the GUI version.

### DIFF
--- a/ykman/logging_setup.py
+++ b/ykman/logging_setup.py
@@ -28,6 +28,7 @@
 from __future__ import absolute_import
 
 import logging
+import ykman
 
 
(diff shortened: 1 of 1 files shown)


dagheyman:
I'm fine with the first part. Shelling out to git to get the version is a bit too smart for my taste, but if you find it useful I can live with it.

dainnilsson:
I agree with Dag. I'm happy to have git specific stuff in packaging scripts and such, but having it as part of the main library code seems a bit weird to me.

emlun:
Ok, I'll just merge the first commit then.

PR #64, opened 2018-02-21
Fix message about skipping tests when serial env var is not set

emlun:
This lets unittest report that the tet was skipped because the
environment variables were not set, instead of falsely "detecting" that
required modes are incorrect.


emlun:
This bug was introduced in c1c92fcbd8d0688570b78a223b14e35a9f842f3b.

PR #66, opened 2018-02-23
Add more PIV tests and fix some corner case bugs

emlun:
Bugs fixed:

- `_pivman_data` was not updated after reset. This could cause subsequent `verify` calls to fail because the `has_stored_key` could remain True in memory although no key remained in the protected data.
- `get_pin_tries()` and `get_puk_tries()` captured the status from an exception assumed to be raised, which is not raised by `get_pin_tries()` if the PIN has already been verified. Instead an empty byte array is returned, which then crashes the calling code (which appears to have been intended to never actually run) attempting to unpack it into two variables. Passing the argument `check=None` instead allows to capture the status reliably.
- `set_mgm_key(None, store_on_device=True)` would overwrite the management key and then fail to write the new random key into protected storage if called without first verifying the PIN. This is fixed by initializing the protected storage before writing the new management key.


emlun:
Review note: `test_piv.py` was renamed to `test_cli_piv.py`, and then `test_piv.py` was created again with new content. GitHub's "Files changed" view confuses this for `test_cli_piv.py` being a new file and `test_piv.py` having a mess of changes. I suggest reviewing commit 9c4b4a2 and [the rest][diff] separately.

[diff]: https://github.com/Yubico/yubikey-manager/pull/66/files/9c4b4a2d1ba7d2c6e75c51d08792e6c6f671c144..4bed9bb997741c967bc16dc7f9acd6df2ad8f47a

test/on_yubikey/test_piv.py
```diff
@@ -0,0 +1,121 @@
...
+@unittest.skipIf(*missing_mode(TRANSPORT.CCID))
+class PivTestCase(DestructiveYubikeyTestCase):
+    pass
+
+
+class ManagementKey(PivTestCase):
+
+    def setUp(self):
+        self.dev = open_device(transports=TRANSPORT.CCID)
+        self.ctrl = PivController(self.dev.driver)
```
dagheyman: Change to `self.controller`. Not because I think the short version is bad, but for consistency.

test/on_yubikey/test_piv.py
```diff
@@ -0,0 +1,121 @@
...
+        self.ctrl.reset()
+
+    def tearDown(self):
```
dagheyman: Should we do a reset in tearDown as well? I noticed the device have a stored mgm key after running all the tests.

ykman/piv.py
```diff
@@ -585,6 +588,17 @@ def set_mgm_key(self, new_key, touch=False, store_on_device=False):
...
+        if store_on_device or (not store_on_device and self.has_stored_key):
+            # Ensure we have access to protected data before overwriting key
+            try:
+                self._init_pivman_protected()
+            except Exception as e:
+                logger.debug('Failed to initialize protected pivman data',
+                             exc_info=e)
+
+                if store_on_device:
+                    raise e
```
dagheyman: Any reason for `raise e` and not just an empty `raise` here ? I think we usually just use `rasie` so you might want to change to that for consistency.

dagheyman:
Nice finds, added some nitpicking review comments.

test/on_yubikey/test_piv.py
emlun: I could add one to `tearDownClass`, which runs once after all the tests in the class. I'm not sure what value it would really add, though. I rather think that if other tests make asumptions about the state of the YubiKey it's probably better to enforce those assumptions in those tests' `setUp` or `setUpClass` instead of relying on other tests to clean up themselves - which most of the time won't be necessary anyway, and it takes nozero time to do the cleanup after all. :)

test/on_yubikey/test_piv.py
emlun: Sure. I didn't think much of the naming since this is just a self-contained test, but that's a very good point.

ykman/piv.py
emlun: No reason other than me not being used to bare `raise` statements being a thing. :)

test/on_yubikey/test_piv.py
dagheyman: Agreed, other tests should not make this kind of assumptions, but instead use their own setUp logic. This was more about having the device as "clean" as possible after all the tests are finished.

emlun:
Thanks @dagheyman!

Lifting my response to [this][thread] out here, because GitHub marked the thread as outdated although it's still unresolved.

I could add one to `tearDownClass`, which runs once after all the tests in the class. I'm not sure what value it would really add, though. I rather think that if other tests make asumptions about the state of the YubiKey it's probably better to enforce those assumptions in those tests' `setUp` or `setUpClass` instead of relying on other tests to clean up themselves - which most of the time won't be necessary anyway, and it takes nozero time to do the cleanup after all. :)


[thread]: https://github.com/Yubico/yubikey-manager/pull/66#discussion_r170510655

test/on_yubikey/test_piv.py
emlun: Yeah, I guess that would be nice. Just one PIV reset costs about 1.5 seconds, though, so I'm a bit skeptical.

PR #73, opened 2018-03-07
Move reusable logic from PIV CLI module to lib module

emlun:
...and also add a bunch of tests for the Python API.


ykman/piv.py
```diff
@@ -853,3 +915,20 @@ def sign_csr_builder(self, slot, public_key, builder, touch_callback=None):
         der = Tlv(0x30, b''.join(seq))
 
         return x509.load_der_x509_csr(der, default_backend())
+
+    @property
+    def supported_pin_policies(self):
```
dagheyman: Any reason not to have this as a boolean "supports_pin_policies" ? It's all or nothing anyway right?

emlun: Yeah, I suppose this one could be simplified like that. My thinking was that it might be possible that more are added in the future (like if YK5 supports some YK4 does not), perhaps? But yeah, might be better to keep it simple if we don't deem that likely.

Perhaps a more compelling argument: This makes it symmetrical to the touch policies, which are already not all or nothing.

I'm not sure which side I prefer...

dagheyman: Both legit points. I think I'd still go with the simplified version because it describes the functionality of the device better (the all or nothing-ness can be implied by the name), but your call.

emlun: I changed it to the simplified version. Thanks!

PR #75, opened 2018-03-09
add support for form factor tag

dagheyman:
- read form factor info if available and save on device
- print out form factor in ykman info.

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -133,8 +133,8 @@ class TestUtilityFunctions(unittest.TestCase):
 
     def test_form_factor_from_code(self):
         self.assertEqual(FORM_FACTOR.UNKNOWN, FORM_FACTOR.from_code(None))
-        self.assertEqual(
-            FORM_FACTOR.UNKNOWN, FORM_FACTOR.from_code('im a string'))
+        with self.assertRaises(ValueError):
+            FORM_FACTOR.from_code('im a string')
         self.assertEqual(FORM_FACTOR.UNKNOWN, FORM_FACTOR.from_code(0x00))
         self.assertEqual(
             FORM_FACTOR.USB_A_KEYCHAIN, FORM_FACTOR.from_code(0x01))
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -3,7 +3,7 @@
 from ykman.util import (b2len, format_code, generate_static_pw,
                         hmac_shorten_key, modhex_decode, modhex_encode,
                         parse_tlvs, parse_truncated, time_challenge, Tlv,
-                        is_pkcs12)
+                        is_pkcs12, FORM_FACTOR)
 from .util import open_file
 import unittest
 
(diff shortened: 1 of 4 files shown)


ykman/util.py
```diff
@@ -76,6 +76,26 @@ def dependent_on_ccid():
...
+@unique
+class FORM_FACTOR(IntEnum):
+    UNKNOWN = 0x00
+    USB_A_KEYCHAIN = 0x01
+    USB_A_NANO = 0x02
+    USB_C_KEYCHAIN = 0x03
+    USB_C_NANO = 0x04
+
+    @classmethod
+    def printable(cls, form_factor):
```
dainnilsson: I would override `__str__(self) ` instead, so that you can just print a value and have it look good, without having to explicitly call printable.

ykman/device.py
```diff
@@ -113,6 +114,8 @@ def _parse_capabilities(self, data):
                 self._serial = int(b2a_hex(tlv.value), 16)
             elif YK4_ENABLED_TAG == tlv.tag:
                 self.enabled = int(b2a_hex(tlv.value), 16)
+            elif YK4_FORMFACTOR_TAG == tlv.tag:
+                self._form_factor = FORM_FACTOR(int(b2a_hex(tlv.value), 16))
```
dainnilsson: The exception where the byte value is not in the `FORM_FACTOR` enum should probably be handled, so that future form factors don't break things. In such a case, `self._form_factor` should probably just be set to the int value.

ykman/device.py
```diff
@@ -61,6 +61,7 @@ def __init__(self, descriptor, driver):
         self._descriptor = descriptor
         self._driver = driver
         self.device_name = descriptor.key_type.value
+        self._form_factor = None
```
dainnilsson: Not sure if it makes sense to differentiate between the case where the formfactor value is missing vs present but 0.

dagheyman:
Tried to make it a bit more future proof now.

test/test_util.py
```diff
@@ -130,3 +130,16 @@ def test_is_pkcs12(self):
             'rsa_2048_key_cert_encrypted.pfx') as \
                 rsa_2048_key_cert_encrypted_pfx:
             self.assertTrue(is_pkcs12(rsa_2048_key_cert_encrypted_pfx.read()))
+
+    def test_form_factor_from_code(self):
+        self.assertEqual(FORM_FACTOR.UNKNOWN, FORM_FACTOR.from_code(None))
+        self.assertEqual(
+            FORM_FACTOR.UNKNOWN, FORM_FACTOR.from_code('im a string'))
```
emlun: I feel like calling `from_code` with a string argument should probably raise an exception instead - because that should never happen, right?

dagheyman: Sure, that's true. I'll update the PR.

ykman/util.py
```diff
@@ -98,10 +98,9 @@ def __str__(self):
 
     @classmethod
     def from_code(cls, code):
-        if code in cls.__members__.values():
-            return cls(code)
-        else:
-            return cls.UNKNOWN
+        if code and not isinstance(code, int):
+            raise ValueError('Invalid form factor code!')
```
emlun: Perhaps the error message could include the invalid value?

PR #76, opened 2018-03-12
refactor scanmap a bit

dagheyman:
- store scancodes in a dictionary instead of in a list. 
- prepare for non-us keyboard layouts. 
- add basic tests.


test/test_scanmap.py
```diff
@@ -0,0 +1,112 @@
...
+        self.assertEqual(b'\xa3', get_scan_codes('^'))
+        self.assertEqual(b'\xad', get_scan_codes('_'))
+        self.assertEqual(b'\xaf', get_scan_codes('{'))
+        self.assertEqual(b'\xb0', get_scan_codes('}'))
+        self.assertEqual(b'\xb2', get_scan_codes('|'))
+        self.assertEqual(b'\xb5', get_scan_codes('~'))
+
+        self.assertEqual(b'\x04\x05\x06', get_scan_codes('abc'))
+        with self.assertRaises(ValueError):
+            get_scan_codes('ööööö')
```
emlun: Is the number of `ö` important? Why five?

ykman/cli/slot.py
```diff
@@ -252,9 +252,9 @@ def static(ctx, slot, password, no_enter, force):
                 else:
                     break
 
-        password = generate_static_pw(chars)
+        password = generate_static_pw(chars).decode()
```
emlun: Why no `'utf-8'` argument?

dagheyman: Because it's the default value anyway. But I can add it if it makes it more clear / for consistency reasons.

test/test_scanmap.py
dagheyman: Good point, updated.

ykman/scanmap/__init__.py
```diff
@@ -0,0 +1,17 @@
+from enum import Enum
```
dainnilsson: Add `from __future__ import absolute_import` to the top.

PR #77, opened 2018-03-14
CLI: updates for static password

dagheyman:
- make the default allowed scancodes the modhex based ones
- add support for german scancodes
- `--force` flag doesn't generate passwords anymore, `--generate` + `--length` is needed for that
- prompt user when overwriting a slot
- rename scanmap module -> scancodes


emlun:
Why is the Python 3.4 build failing?

PR #79, opened 2018-03-19
Add -g|--generate flags to misc. commands

emlun:
- `ykman piv change-management-key`:
  - PIV commands called with `-f|--force` will now crash instead of interactively prompting for PIN or management key.
  - `-f|--force` variants without `-p|--protect` or `-g|--generate` will now crash instead of generating a random management key.
  - ~~Interactive prompt variants still generate a random management key on blank input, even without `-g|--generate`.~~
  - Interactive prompt variants no longer generate a random management key on blank input.
- `ykman slot chalresp`
- `ykman slot otp`


emlun:
s/crash/fail/

ykman/cli/piv.py
```diff
@@ -648,26 +659,41 @@ def change_management_key(
...
+                click.echo(
+                    'Generated management key: {}'.format(
+                        b2a_hex(new_management_key).decode('utf-8')))
+
+        elif force:
+            ctx.fail('New management key not given. Please remove the --force '
+                     'flag, or set the --generate flag or the '
+                     '--new-management-key option.')
+
+        else:
```
dagheyman: So for the static password command we actually dropped the "blank to randomly generate" functionality, I'd suggest we do the same thing here to simplify the code and have similar behaviour.

emlun: Sounds good to me!

emlun:
@dagheyman I updated the `ykman slot chalresp` and `ykman slot otp` commands as well, if you want to review them too.

ykman/cli/slot.py
```diff
@@ -170,57 +170,90 @@ def delete(ctx, slot, force):
               callback=parse_hex(16), help='16 byte secret key, in hex.')
 @click.option('--no-enter', is_flag=True, help="Don't send an Enter "
               'keystroke after outputting an OTP.')
+@click.option(
```
dagheyman: Do we need the alias command? We don't have that anywhere else. I'd prefer coming up with short flags for the below flags instead.

emlun: Yeah, good point. The main reason I added it was actually as a way to decouple the [`test_update_settings_enter_slot_2`][test1] and [`test_delete_slot_2`][test2] tests from the internal logic a bit, but I think I could as well make them use a static password slot instead.

[test1]: https://github.com/Yubico/yubikey-manager/pull/79/files/e1ec2ed89e94ae95505a9342446ab8a75f5243c3#diff-c943060aa5353f04a2721936c20d2a33L174
[test2]: https://github.com/Yubico/yubikey-manager/pull/79/files/e1ec2ed89e94ae95505a9342446ab8a75f5243c3#diff-c943060aa5353f04a2721936c20d2a33L179

emlun: Fixed.

emlun:
Added short variants of the generate flags to `ykman slot otp`: `-S|--serial-public-id` (capital to mirror `-P`), `-g|--generate-private-id`, `-G|--generate-key`.

PR #81, opened 2018-03-20
Add FIDO 2 functionality

dagheyman:
- drop dependency on libu2f-host
- add dependency on fido-host
- add fido commands
- add support for fido capability


dagheyman:
I'd like a second look at this one @dainnilsson

PR #82, opened 2018-03-20
Fix PIV cert generation when sign requires PIN

emlun:
Commit 5003df366637cfaac80ab9b8c987aa402d9bf20e exposed a bug where generating self-signed PIV certificates and CSRs for slot `9c` would fail. This is because the `9c` slot has default pin policy `ALWAYS`.

These commits

- add a `pin=None` parameter to the `PivController.sign()` method.
  - If the signing operation fails with status `6a82` (Access denied) and a `pin` argument was given, the `sign()` method will attempt to verify that PIN and recursively retry itself.
- add a `pin=None` parameter to the following methods of `PivController`, which forward the `pin` argument in the internal call to `sign()`:
  - `generate_certificate_signing_request()`
  - `generate_self_signed_certificate()`
  - `sign_cert_builder()`
  - `sign_csr_builder()`
- change some uses of the `sign()` method to send the `pin` argument to `sign()` instead of calling the `verify()` method and then `sign()` without a `pin` argument.


test/on_yubikey/test_cli_piv.py
```diff
@@ -167,6 +152,144 @@ def test_export_attestation_certificate(self):
         self.assertIn('BEGIN CERTIFICATE', output)
 
 
+class GenerateSelfSigned_NonDefaultMgmKey(PivTestCase):
```
dagheyman: Not a fan of underscore in classnames.

emlun: I agree, but I think it makes a useful visual separator in this case. Do you prefer this

```
test_generate_self_signed_slot_9a (test.on_yubikey.test_cli_piv.GenerateSelfSigned_NonDefaultMgmKey) ... ok
test_generate_self_signed_slot_9c (test.on_yubikey.test_cli_piv.GenerateSelfSigned_NonDefaultMgmKey) ... ok
test_generate_self_signed_slot_9d (test.on_yubikey.test_cli_piv.GenerateSelfSigned_NonDefaultMgmKey) ... ok
test_generate_self_signed_slot_9e (test.on_yubikey.test_cli_piv.GenerateSelfSigned_NonDefaultMgmKey) ... ok
test_generate_self_signed_slot_9a (test.on_yubikey.test_cli_piv.GenerateSelfSigned_ProtectedMgmKey) ... ok
test_generate_self_signed_slot_9c (test.on_yubikey.test_cli_piv.GenerateSelfSigned_ProtectedMgmKey) ... ok
test_generate_self_signed_slot_9d (test.on_yubikey.test_cli_piv.GenerateSelfSigned_ProtectedMgmKey) ... ok
test_generate_self_signed_slot_9e (test.on_yubikey.test_cli_piv.GenerateSelfSigned_ProtectedMgmKey) ... ok
```
or this?
```
test_generate_self_signed_slot_9a (test.on_yubikey.test_cli_piv.GenerateSelfSignedWithNonDefaultMgmKey) ... ok
test_generate_self_signed_slot_9c (test.on_yubikey.test_cli_piv.GenerateSelfSignedWithNonDefaultMgmKey) ... ok
test_generate_self_signed_slot_9d (test.on_yubikey.test_cli_piv.GenerateSelfSignedWithNonDefaultMgmKey) ... ok
test_generate_self_signed_slot_9e (test.on_yubikey.test_cli_piv.GenerateSelfSignedWithNonDefaultMgmKey) ... ok
test_generate_self_signed_slot_9a (test.on_yubikey.test_cli_piv.GenerateSelfSignedWithProtectedMgmKey) ... ok
test_generate_self_signed_slot_9c (test.on_yubikey.test_cli_piv.GenerateSelfSignedWithProtectedMgmKey) ... ok
test_generate_self_signed_slot_9d (test.on_yubikey.test_cli_piv.GenerateSelfSignedWithProtectedMgmKey) ... ok
test_generate_self_signed_slot_9e (test.on_yubikey.test_cli_piv.GenerateSelfSignedWithProtectedMgmKey) ... ok
```

...and it's "just" the tests, not part of any API. :)

dagheyman: I prefer the second version just because my brain reacts to the mixing of styles, but don't have very strong opinions.

emlun: Perhaps I should move those tests to a separate module instead...

emlun:
@dainnilsson if you want to review this:

- [Changes before moving files around](https://github.com/Yubico/yubikey-manager/pull/82/files/efe84f19c617e7db886ac51aa3aa848ccd2be064)
- [Changes moving files around](https://github.com/Yubico/yubikey-manager/pull/82/files/efe84f19c617e7db886ac51aa3aa848ccd2be064..65129a83805aca8ad926424f2a0089e3de81c185)
- [Changes after moving files around](https://github.com/Yubico/yubikey-manager/pull/82/files/65129a83805aca8ad926424f2a0089e3de81c185..cdeaadf217a43218189fd6d414b18d06eb62f0c0)

ykman/piv.py
```diff
@@ -814,14 +815,27 @@ def sign_raw(self, slot, algorithm, message):
...
-            message = h.finalize()
-        return self.sign_raw(slot, algorithm, message)
+            message_hash = h.finalize()
+        try:
+            return self.sign_raw(slot, algorithm, message_hash)
+        except APDUError as e:
+            if e.sw == SW.ACCESS_DENIED:
+                if pin:
+                    self.verify(pin, touch_callback=touch_callback)
+                    return self.sign(slot, algorithm, message, pin=pin)
```
dainnilsson: Isn't it wrong to pass the pin again to the recursive call? If it didn't work the first time...

ykman/piv.py
```diff
@@ -814,14 +815,27 @@ def sign_raw(self, slot, algorithm, message):
...
+                else:
```
dainnilsson: You can get rid of this else-clause and just de-indent the raise at the bottom of the method by one level.

ykman/piv.py
emlun: No, in fact it's required because - as you note above - the VERIFY command might follow up with an AUTHENTICATE command which consumes the verifiedness needed for the ALWAYS pin policy. If the PIN is wrong, then the `verify` call here will instead raise an exception that isn't caught here.

ykman/piv.py
emlun: Yes. Is that the require style in this codebase? Personally I generally prefer explicit branches over early returns.

ykman/piv.py
dainnilsson: So, if I'm understanding this correctly, the first call may get SW.ACCESS_DENIED due to lack of "verifiedness", which then calls verify(), and then a recursive call to sign, again passing pin.
However, the call to verify() might not provide the "verifiedness" the first time, as it might also do an authenticate(), which results in a second SW.ACCESS_DENIED. In that case a second recursive call to sign() is needed, making it a total of 3 invocations of sign() before success?

What about the `touch_callback`? That seems to get lost between the first and second calls to verify().

emlun:
@dainnilsson 
>Is the reason for adding the `pin` argument to sign, rather than explicitly calling verify() before calling sign(), due to verify() also calling authenticate() in some circumstances (making the verification unusable for a PIN policy of ALWAYS)?

Yes, that's the main reason. I also thought it would make sense that if signing might require PIN verification, then the sign method could accept an optional PIN parameter.

>Would having PivController.verify() call the actual VERIFY instruction again at the end, if it called authenticate(), solve the problem?

Yes, that is an alternative (and much smaller) solution. I think the reason why I dind't go with that one is because I came at the problem from the other direction and discovered this after the fact.

>I think it makes the API a bit weird having sign take a pin argument which it doesn't use unless it gets an error. This means that the call may succeed even if the PIN is wrong.

This is a very good point. I agree, then, that staying with separate calls to `verify()` and `sign()` is the way to go.

>Generally I would prefer to catch the error at a higher level, and THEN prompt the user for the PIN in the case where it's needed, rather than preemptively doing so.

I'll note that you can still do this by not sending a `pin` argument to `sign()`, and then prompting for PIN and retrying, but it does kind of defeat the purpose of the `pin` parameter.

dainnilsson:
That makes sense. How about this:
- Modify verify() so that it always provides "verifiedness" (which may result in multiple calls to VERIFY), since it's weird that it doesn't currently do so (my bad!).

And if you think it makes sense to do so, also:
- Keep the optional `pin` argument for the `sign_X_builder()` and `generate_X()` methods, and if provided, verify() WILL be called prior to sign(), causing an error if the PIN is wrong. If not provided, then the call may result in ACCESS_DENIED.

ykman/piv.py
dagheyman: I prefer to avoid else branches [if possible].(https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement?answertab=votes#tab-top)

emlun:
Actually, thinking back at this again... I'd say the main issue here is that there's no way (I'm told) to query for the PIN policy of the key slot to use. That didn't seem like something the caller should have to care about, which is why I integrated the logic in `sign()`. But I realize now that the `pin` parameter doesn't actually help much with that - if the PIN policy _might_ be ALWAYS, then to be sure you'd have to pass a `pin` argument even if the PIN policy is in fact NEVER, so the caller still has to handle the failure externally if it wants to save the user from having to provide a PIN when none is needed.

So I think it's probably best to revert the addition of the `pin` parameter and just fix `verify()` instead. That also means there won't be two ways to do this.

dainnilsson:
Yeah, that makes sense. I agree.

ykman/piv.py
emlun: Yes, that is a correct understanding.

>What about the `touch_callback`?

Oops, looks like I forgot about that one...

emlun:
#83 replaces this.

PR #85, opened 2018-03-23
make fido reset more interactive

dagheyman:


ykman/cli/fido.py
```diff
@@ -158,16 +159,20 @@ def reset(ctx):
...
+    click.echo('Remove and re-insert your YubiKey to perform the reset...')
+
+    removed = False
+    while True:
+        descriptors = list(get_descriptors())
+        n_keys = len(descriptors)
+        if not n_keys:
+            removed = True
+        if removed and n_keys == 1:
+            break
```
emlun: How about a short sleep here (~50 ms or so), so we don't max the processor for this?

PR #92, opened 2018-04-09
Add prettier names for a couple of capabilities

dagheyman:


dagheyman:
> I feel like it might be a tiny bit cleaner to do this by lookup in a dictionary instead, but that might be overkill in this case.

I followed the style of how it's done for the FORM_FACTOR enum in the same file.

emlun:
Good point. :)

PR #99, opened 2018-04-20
Rename U2FDriver -> FidoDriver.

dainnilsson:


ykman/descriptor.py
```diff
@@ -30,7 +30,7 @@
 from .util import PID, TRANSPORT, Mode
 from .device import YubiKey
 from .driver_ccid import open_devices as open_ccid
-from .driver_u2f import open_devices as open_u2f
+from .driver_fido import open_devices as open_u2f
```
dagheyman: Should be imported as open_fido

dainnilsson: Done

PR #103, opened 2018-04-25
Fix destructive tests in Python 2

emlun:
- Remove unused type conversions that didn't work in Python 2
- Pass `u''` strings to make `x509.NameAttribute` happy in Python 2
- Don't use `assertRaisesRegex` which was added in Python 3


emlun:
Much better indeed - thanks!

PR #104, opened 2018-04-30
Make all non-short options case insensitive

dagheyman:
The idea is that all options that are not short-options should be case insensitive. This is done using the [token_normalize_func](http://click.pocoo.org/6/api/#click.Context.token_normalize_func), as suggested in [this thread](https://github.com/pallets/click/issues/569).

The alternative is to normalize all options (not only those with a >1 length), but then we can't have options like `-p` and `-P` have different meanings.

### DIFF
--- a/test/on_yubikey/test_cli_config.py
+++ b/test/on_yubikey/test_cli_config.py
@@ -138,6 +138,11 @@ class TestConfigNFC(DestructiveYubikeyTestCase):
             ykman_cli(
                 'config', 'nfc', '--enable-all', '--disable-all', 'FIDO2', '-f')
 
+    def test_case_insensitive(self):
+        ykman_cli('config', 'nfc', '--LIST')  # Assume no exception
+        with self.assertRaises(SystemExit):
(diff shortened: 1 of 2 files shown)


emlun:
Why should they be case insensitive?

dagheyman:
> Why should they be case insensitive?

UX reasons.
1. It's faster to type commands when they are case insensitive.
2. Users will probably try the lowercase version first, and then get an error, having to re-type the whole command.

emlun:
But we don't have any uppercase `--long-options`, do we? I agree this is a good thing for option _values_ and that kind of thing, such as the mode/application names (`CCID`, `FIDO2` etc) in the `mode`/`config` commands, but I'm not sure about doing the same for the option _names_. Sure, we won't have long options whose names differ only in case - but if the long options are case insensitive, might that make some think the short options are as well?

dagheyman:
I see your point, and agree, will look into alternative solutions that only affect option _vaues_.

PR #108, opened 2018-05-07
address issue 31: support '--json' output

blaisep:
@dagheyman Feel free to use this PR if you think it will help the request for JSON output.


dagheyman:
Thanks for the PR! It looks fine, however we decided we will not add json support for this tool, and point to the python API for scripting purposes.

PR #109, opened 2018-05-15
otp: try reading config twice if timeout

dagheyman:
Using the OTP interface directly after using FIDO might result in a timeout. If this happens, wait 3 seconds and try again.

To reproduce the error that this fixes:

```
from ykman import driver_fido, driver_otp
next(driver_fido.open_devices()).read_config()
next(driver_otp.open_devices()).read_config()
```


emlun:
Can that snippet be added as a test?

dagheyman:
> Can that snippet be added as a test?
Should be possible, good point.

PR #118, opened 2018-06-13
Don't validate old PIN in change_pin

emlun:
The old PIN is by definition already set, so there's no point in validating it.

This change also lets us reuse this code for setting the initial PIN for FIPS U2F - since there's only one SET_PIN command (for both change and initial set) and there's no command to check if a PIN is already set, we'll simply use change_pin and pass an empty old PIN in that case. This removed validation would prevent passing an empty old PIN.

### DIFF
--- a/ykman/cli/fido.py
+++ b/ykman/cli/fido.py
@@ -101,7 +101,6 @@ def set_pin(ctx, pin, new_pin):
                     show_default=False)
 
     def change_pin(pin, new_pin):
-        fail_if_not_valid(ctx, pin)
         fail_if_not_valid(ctx, new_pin)
         try:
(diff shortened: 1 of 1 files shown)


dagheyman:
> The old PIN is by definition already set, so there's no point in validating it.

I guess the reasoning here is that if the old PIN is invalid, it is a user mistake and we don't need to waste a PIN retry by trying to use it?

emlun:
Ah, hm, that is a good point. Then, how about adding a parameter that controls whether an empty current PIN is acceptable (which defaults to not)?

emlun:
Actually, doing that won't change any existing behaviour, so I'll bundle that in with the other FIPS support changes.

PR #121, opened 2018-06-19
Refactorisations in preparation of adding FIPS support

emlun:
- Allow empty current pin in internal change_pin function. The FIPS U2F PIN has only one command to both set and change PIN, and there's no good way to tell if a PIN is already set, so this is needed for initially setting the U2F PIN.
- Extract controller-polymorphic try_reset function in fido reset command


dagheyman:
LGTM

PR #122, opened 2018-06-19
Add ability to set OTP access code in ykman otp settings

emlun:
This will be needed for FIPS support.

### DIFF
--- a/ykman/cli/otp.py
+++ b/ykman/cli/otp.py
@@ -495,8 +495,9 @@ def hotp(ctx, slot, key, digits, counter, no_enter, force):
 @click_force_option
 @click.pass_context
 @click.option(
-    '-A', '--new-access-code', metavar='HEX',
-    help='Set a new access code for the slot.')
+    '-A', '--new-access-code', metavar='HEX', required=False,
+    help='Set a new 6 byte access code for the slot. Set to empty to use a '
+         'prompt for input.')
 @click.option(
     '--delete-access-code', is_flag=True,
     help='Remove access code from the slot.')
@@ -524,6 +525,10 @@ def settings(ctx, slot, new_access_code, delete_access_code, enter, pacing,
         ctx.fail('Not possible to update settings on an empty slot.')
 
     if new_access_code is not None:
+        if new_access_code == '':
+            new_access_code = click.prompt(
+                'Enter new access code', show_default=False)
+
         try:
             new_access_code = parse_access_code_hex(new_access_code)
         except Exception as e:
--- a/ykman/cli/otp.py
+++ b/ykman/cli/otp.py
@@ -517,13 +517,13 @@ def settings(ctx, slot, new_access_code, delete_access_code, enter, pacing,
     """
     controller = ctx.obj['controller']
 
-    if new_access_code and delete_access_code:
+    if (new_access_code is not None) and delete_access_code:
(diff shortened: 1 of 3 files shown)


ykman/cli/otp.py
```diff
@@ -524,5 +534,8 @@ def settings(ctx, slot, enter, pacing, force):
     except YkpersError as e:
         _failed_to_write_msg(ctx, e)
 
+    if new_access_code is not None:
```
dagheyman: Any reason not to do `if new_access_code:` instead of `if new_access_code is not None:` ?

emlun: Not really... It would be if `new_access_code` could be empty, but that's not a valid value, so `if new_access_code:` should be functionally equivalent in this case. I'll change it.

emlun: Actually, I realized only one of the two occurrences can be changed. If the first is `if new_access_code:`, then the parsing step is never entered, the program accepts the input and does nothing.

dagheyman:
These are failing for me on a YK4 (non FIPS)
```
test_new_access_code_conflicts_with_delete_access_code (test.on_yubikey.test_cli_otp.TestSlotProgramming) ... FAIL
test_set_access_code_prompt_slot_2 (test.on_yubikey.test_cli_otp.TestSlotProgramming) ... FAIL
test_set_access_code_slot_2 (test.on_yubikey.test_cli_otp.TestSlotProgramming) ... FAIL
```

```
======================================================================
FAIL: test_new_access_code_conflicts_with_delete_access_code (test.on_yubikey.test_cli_otp.TestSlotProgramming)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/user/work/code/yubikey-manager/test/on_yubikey/test_cli_otp.py", line 320, in test_new_access_code_conflicts_with_delete_access_code
    self._check_slot_2_has_access_code()
  File "/Users/user/work/code/yubikey-manager/test/on_yubikey/test_cli_otp.py", line 335, in _check_slot_2_has_access_code
    '2', '-f')
AssertionError: SystemExit not raised

======================================================================
FAIL: test_set_access_code_prompt_slot_2 (test.on_yubikey.test_cli_otp.TestSlotProgramming)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/user/work/code/yubikey-manager/test/on_yubikey/test_cli_otp.py", line 295, in test_set_access_code_prompt_slot_2
    self._check_slot_2_has_access_code()
  File "/Users/user/work/code/yubikey-manager/test/on_yubikey/test_cli_otp.py", line 335, in _check_slot_2_has_access_code
    '2', '-f')
AssertionError: SystemExit not raised

======================================================================
FAIL: test_set_access_code_slot_2 (test.on_yubikey.test_cli_otp.TestSlotProgramming)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/user/work/code/yubikey-manager/test/on_yubikey/test_cli_otp.py", line 279, in test_set_access_code_slot_2
    self._check_slot_2_has_access_code()
  File "/Users/user/work/code/yubikey-manager/test/on_yubikey/test_cli_otp.py", line 335, in _check_slot_2_has_access_code
    '2', '-f')
AssertionError: SystemExit not raise
```

emlun:
Good catch, thanks! Fixed in #126.

PR #123, opened 2018-06-21
Add FIPS support

emlun:
Points to perhaps pay extra attention to:

- The text output in the [root `info` command](https://github.com/Yubico/yubikey-manager/compare/fips?expand=1#diff-7bb2e04e733e115dd9f9e83f473d8e13R55)
- The version numbers that [identify FIPS keys](https://github.com/Yubico/yubikey-manager/compare/fips?expand=1#diff-f9537064cbd20a47ccb76d21f7301286R315)


ykman/cli/fido.py
```diff
@@ -46,11 +50,14 @@ def fido(ctx):
     """
     Manage FIDO applications.
     """
-    try:
-        ctx.obj['controller'] = Fido2Controller(ctx.obj['dev'].driver)
-    except Exception as e:
-        logger.debug('Failed to load Fido2Controller', exc_info=e)
-        ctx.fail('Failed to load FIDO 2 Application.')
+    if ctx.obj['dev'].is_fips:
+        ctx.obj['controller'] = FipsU2fController(ctx.obj['dev'].driver)
```
dagheyman: Can this fail too? Maybe do a generic try-catch around the whole thing?

ykman/cli/fido.py
```diff
@@ -176,23 +219,54 @@ def try_reset(controller_type):
...
-                'Reset failed. You need to touch your'
-                ' YubiKey to confirm the reset.')
-        elif e.code == CtapError.ERR.NOT_ALLOWED:
-            ctx.fail(
-                'Reset failed. Reset must be triggered within 5 seconds'
-                ' after the YubiKey is inserted.')
-        else:
+    if ctx.obj['dev'].is_fips:
+        if not force:
+            click.echo(
```
dagheyman: This `click.echo` could me merged with the `click.prompt` below, no?

ykman/cli/fido.py
```diff
@@ -176,23 +219,54 @@ def try_reset(controller_type):
...
+                '\nWARNING: This will destroy the FIPS attestation key, making '
+                'this YubiKey permanently incapable of FIPS mode.'
+            )
+            destroy_input = click.prompt(
```
dagheyman: I don't like this, we don't do it anywhere else and there is a big warning anyway. I suggest a regular prompt. If the user is doing `fido reset` from the command line on a FIPS key (not a consumer device), they hopefully know what they are doing.

ykman/cli/info.py
```diff
@@ -50,6 +50,10 @@ def info(ctx):
     else:
         click.echo('Firmware version: Uncertain, re-run with only one '
                    'YubiKey connected')
+
+    if dev.is_fips:
+        click.echo('FIPS capable: Yes')
```
dagheyman: Nitpick: Maybe print out something like `This YubiKey is capable of FIPS mode.`, since it can never be `FIPS capable: No`'

ykman/cli/otp.py
```diff
@@ -131,6 +131,24 @@ def info(ctx):
     controller = ctx.obj['controller']
     click.echo(dev.device_name)
     slot1, slot2 = controller.slot_status
+
+    if dev.is_fips:
+        fips_mode = controller.is_in_fips_mode
+        summary = fips_mode and 'Yes' or 'No'
```
dagheyman: Is `Yes/No` a summary? Maybe this will be easier to follow if click.echo is called directly if fips mode, and the different explanations in the else branch?

ykman/fido.py
```diff
@@ -63,3 +68,67 @@ def reset(self, touch_callback=None):
...
+        # We don't know, but the change and set commands are the same here.
+        return True
+
+    def set_pin(self, pin):
+        raise NotImplementedError('Use the change_pin method instead.')
+
+    def change_pin(self, old_pin, new_pin):
+        new_length = len(new_pin)
+
+        old_pin = old_pin.encode('utf-8')
```
dagheyman: .encode() is fine

ykman/fido.py
```diff
@@ -63,3 +68,67 @@ def reset(self, touch_callback=None):
...
+    def is_fips(self):
+        return True
+
+    @property
+    def is_in_fips_mode(self):
+        try:
+            self.ctap.send_apdu(ins=FIPS_U2F_CMD.VERIFY_FIPS_MODE)
+            return True
+        except ApduError as e:
+            print('%x' % e.code)
```
dagheyman: drop print or use logger.error/logger.debug

ykman/cli/fido.py
emlun: Good catch, thanks!

ykman/cli/fido.py
emlun: Sure.

ykman/cli/fido.py
emlun: Good points... maybe we should ask TPM about this. I wouldn't be so sure that it's clear that resetting is irreversible (in the sense of resetting the attestation key).

ykman/cli/info.py
emlun: Sure!

ykman/cli/otp.py
emlun: I think my main reason for using only one echo call was to keep it all on one line of output. But I agree the clarity here can be improved, I have some ideas for that.

ykman/fido.py
emlun: Good catch, thanks!

ykman/fido.py
emlun: Decided in F2F discussion to keep this as is.

test/on_yubikey/test_fips_u2f_commands.py
```diff
@@ -0,0 +1,77 @@
...
+
+@unittest.skipIf(not is_fips(), 'FIPS YubiKey required.')
+class TestFipsU2fCommands(DestructiveYubikeyTestCase):
+
+    def test_echo_command(self):
+        dev = open_device(transports=TRANSPORT.FIDO)
+
+        res = dev.driver._dev.call(
+            CTAPHID.MSG,
+            [*[0, FIPS_U2F_CMD.ECHO], 0, 0, *[0, 0, 6], *b'012345'])
```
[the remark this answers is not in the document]
emlun: Oops, thanks!

ykman/cli/otp.py
emlun: Potential solution proposed in https://github.com/Yubico/yubikey-manager/pull/125

emlun: Fixed in #125

dagheyman:
Added a couple of commits to use the terms 'YubiKey FIPS' and 'FIPS Approved Mode'.

dagheyman:
- U2F pin minimum length is 6, currently assuming it's 4
- fido set-pin currently prompts for "Current PIN" first

dagheyman:
- if device is not in [OTP+CCID+FIDO] only mode, `ykman info` crashes

emlun:
`ykman fido set-pin --u2f` will now ask for the current PIN _after_ the new PIN, and only if it needs to. It turns out that failing to set the PIN does not cost a PIN retry if attempted with empty current PIN.

dagheyman:
@emlun 

I dropped all the explanations from the `info` commands, see commit message. Easy to re-add if the feature is requested, but let's start minimal.

I feel pretty comfortable merging this now, to allow a RC1 release, do you feel the same?

dagheyman:
maybe @dain could do a sanity-check review of the PR as a whole as well

dagheyman:
Maybe @dainnilsson can do a sanity check review of the PR as well?

emlun:
Looks good to me!

ykman/device.py
```diff
@@ -308,6 +310,10 @@ def transport(self):
     def mode(self):
         return self._driver.mode
 
+    @property
+    def is_fips(self):
+        return self.version >= (4, 4, 0) and self.version < (4, 5, 0)
```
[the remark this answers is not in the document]
emlun: Done!

PR #125, opened 2018-06-25
Extract function explain_fips_status

emlun:
Potential solution to review comment https://github.com/Yubico/yubikey-manager/pull/123#discussion_r197691547

### DIFF
--- a/ykman/cli/otp.py
+++ b/ykman/cli/otp.py
@@ -89,6 +89,25 @@ def _confirm_slot_overwrite(controller, slot):
             abort=True)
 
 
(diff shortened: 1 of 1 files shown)


ykman/cli/otp.py
```diff
@@ -89,6 +89,25 @@ def _confirm_slot_overwrite(controller, slot):
             abort=True)
 
 
+def explain_fips_status(is_fips_mode, slot1_programmed, slot2_programmed):
```
dagheyman: Looks good, but should maybe be a private method?

emlun: :+1:

PR #128, opened 2018-06-27
Add overall FIPS status to ykman info

emlun:
This would merge into #123.

Does the output format look reasonable?

### DIFF
--- a/ykman/cli/info.py
+++ b/ykman/cli/info.py
@@ -92,7 +92,7 @@ def info(ctx):
         status_keys = list(fips_status.keys())
         status_keys.sort()
         for status_key in status_keys:
-            click.echo('  - {}: {}'.format(
+            click.echo('  {}: {}'.format(
                 status_key, 'Yes' if fips_status[status_key] else 'No'))
 
         click.echo()
--- a/ykman/cli/info.py
+++ b/ykman/cli/info.py
@@ -95,7 +95,7 @@ def info(ctx):
             click.echo('  - {}: {}'.format(
                 status_key, 'Yes' if fips_status[status_key] else 'No'))
 
-    click.echo()
+        click.echo()
 
     rows = []
     for app in APPLICATION:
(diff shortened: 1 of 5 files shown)


ykman/cli/info.py
```diff
@@ -27,10 +27,29 @@
...
+from ..descriptor import open_device
+from ..fido import FipsU2fController
+from ..oath import OathController
+from ..otp import OtpController
+from ..util import APPLICATION, TRANSPORT
 import click
 
 
+def get_overall_fips_status(serial):
+    stati = {}
```
dagheyman: `stati` ?

ykman/cli/info.py
```diff
@@ -64,6 +80,23 @@ def info(ctx):
...
+
+        fips_status = get_overall_fips_status(dev.serial)
+
+        click.echo('FIPS mode active: {}'.format(
+            'Yes' if all(fips_status.values()) else 'No'))
+
+        status_keys = list(fips_status.keys())
+        status_keys.sort()
+        for status_key in status_keys:
+            click.echo('  - {}: {}'.format(
```
dagheyman: I don't like the `-` (we dont use them in other places), let's just use tabs for indentation.

ykman/cli/info.py
```diff
@@ -64,6 +80,23 @@ def info(ctx):
...
+                status_key, 'Yes' if fips_status[status_key] else 'No'))
+
+    click.echo()
```
dagheyman: This one is just need if fips, leads to extra blank line on regular devices.

ykman/cli/info.py
emlun: Meh, seems like [my latin fu failed me](https://english.stackexchange.com/questions/877/what-is-the-plural-form-of-status).

ykman/cli/info.py
emlun: Good catch, thanks!

ykman/cli/info.py
emlun: Fixed!

emlun:
Travis failure is due to a flake8 issue which is fixed in 6799a1753c1d1b49ecf60d0a9ab3dd3f50479ecd in #123; merging.

PR #129, opened 2018-06-28
piv: block use of RSA1024 keys on YubiKey FIPS

dagheyman:

PR #130, opened 2018-06-28
pin policy: YubiKey FIPS does not support NEVER

dagheyman:


dagheyman:
Added a test, see #129 for answer on library vs cli

PR #133, opened 2018-07-05
FIDO: add command to unlock YubiKey FIPS if a PIN is set

dagheyman:


dagheyman:
Perfect, thanks for the additions.

PR #138, opened 2018-08-07
Fix duplicated --force option on fido reset

emlun:


ykman/cli/fido.py
```diff
@@ -186,12 +186,9 @@ def set_pin(new_pin):
         set_pin(new_pin)
 
 
-@click_force_option
 @fido.command('reset')
-@click.confirmation_option(
-            '-f', '--force', prompt='WARNING! This will delete '
-            'all FIDO credentials, including FIDO U2F credentials,'
-            ' and restore factory settings. Proceed?')
+@click.option('-f', '--force', is_flag=True,
```
dagheyman: Why not use `click_force_option` from util.py?

emlun: I think it was because that provided no good way to print the warning message below.

PR #139, opened 2018-08-11
Update development.adoc

ndowens:
Add Howto install on FreeBSD

### DIFF
--- a/doc/development.adoc
+++ b/doc/development.adoc
@@ -16,6 +16,9 @@ and https://developers.yubico.com/yubikey-personalization/[ykpers] DLLs to the r
 
     $ sudo apt install swig libykpers-1-1 libu2f-udev pcscd libpcsclite-dev
 
+==== FreeBSD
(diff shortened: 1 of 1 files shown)


doc/development.adoc
```diff
@@ -16,6 +16,9 @@ and https://developers.yubico.com/yubikey-personalization/[ykpers] DLLs to the r
 
     $ sudo apt install swig libykpers-1-1 libu2f-udev pcscd libpcsclite-dev
 
+==== FreeBSD
+    $ sudo pkg-install security/py-yubikey-manager
```
rickard-von-essen: $ sudo pkg install security/py-yubikey-manager

dagheyman:
While we appreciate the packaging work being done in the community, I don't think we should include installation instructions for platforms we don't support and maintain ourselves. That is currently a source release, pypi, Ubuntu, macOS and Windows.

PR #142, opened 2018-08-23
ndef: allow setting the prefix

dagheyman:


ykman/cli/otp.py
```diff
@@ -158,15 +158,27 @@ def swap(ctx):
 @otp.command()
 @click_slot_argument
 @click.pass_context
-def ndef(ctx, slot):
+@click.option(
+    '-p', '--prefix', help='Added before the NDEF payload. Typically an URI.')
```
emlun: Teeny tiny nit: "a URI" :)

PR #146, opened 2018-09-05
Build stand alone executable

dagheyman:
Use PyInstaller to build ykman as a single executable on macOS and Windows.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,14 @@ matrix:
           sudo: required
           python: 3.6
           services: docker
+          env: DEPLOY=true
+        - os: linux
+          sudo: required
+          python: 2.7
         - os: osx
           sudo: required
           language: generic
-          env: PY_VERSION="3.6.6"
+          env: PY_VERSION="3.6.6" DEPLOY=true
           osx_image: xcode7.3
 
 before_install:
@@ -82,3 +86,4 @@ deploy:
   upload-dir: "yubikey-manager"
   on:
     all_branches: true
+    condition: $DEPLOY = true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,39 +1,74 @@
 language: python
-dist: trusty
-sudo: required
 
-services:
-  - docker
-
-python:
-  - "2.7"
-  - "3.4"
-  - "3.5"
-  - "3.6"
-
-cache:
-  directories:
-    - $HOME/.cache/pip
+matrix:
+    include:
+        - os: linux
+          sudo: required
+          python: 3.6
+          services: docker
+        - os: osx
+          sudo: required
+          language: generic
+          env: PY_VERSION="3.6.6"
+          osx_image: xcode7.3
 
 before_install:
-  - docker build -t xenial -f docker/xenial/Dockerfile .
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,39 +1,74 @@
...
-  - "3.6"
-
-cache:
-  directories:
-    - $HOME/.cache/pip
+matrix:
+    include:
+        - os: linux
+          sudo: required
+          python: 3.6
```
emlun: We should probably keep building with Python 2 while we support it, no?

.travis.yml
```diff
@@ -1,39 +1,74 @@
...
+    fi
 
 install:
-  - pip install pre-commit
-  - pip install -e .
+  - |
+    if [[ "$TRAVIS_OS_NAME" == "linux" ]]
+    then
+      pip install pre-commit
+      pre-commit run --all-files
```
emlun: Should `pre-commit run` not be in `script:` instead?

.travis.yml
```diff
@@ -1,39 +1,74 @@
...
+      pip install -e .
+      python setup.py test
+    fi
 
 script:
-  - pre-commit run --all-files
-  - python setup.py test
```
emlun: Looks like the `python setup.py test` step was lost.

appveyor.yml
```diff
@@ -5,39 +5,59 @@
...
 environment:
-  LIBU2F_HOST_VERSION: "1.1.4"
   YKPERS_VERSION: "1.19.0"
   LIBUSB_VERSION: "1.0.21"
+  YKPERS_BASEURL: https://developers.yubico.com/yubikey-personalization/Releases
 
   matrix:
-    - PYTHON: "C:\\Python27"
-    - PYTHON: "C:\\Python35"
-    - PYTHON: "C:\\Python36"
```
emlun: Again, I think we'd getter keep CI testing Python 2 for now...

appveyor.yml
```diff
@@ -5,39 +5,59 @@
...
+  - ps: Copy-Item ".\libusb\$env:LIBUSB_DLL_FOLDER\dll\*.dll" "." -Force -verbose
 
 build_script:
-  - pip install -e .
-  - ykman -v
-
-test_script:
-  - python setup.py test
+  - pip install .
+  - pip install pyinstaller
```
emlun: Could `pip install pyinstaller` be moved to `install:`?

appveyor.yml
```diff
@@ -5,39 +5,59 @@
...
   - 7z x libusb-%LIBUSB_VERSION%.7z -o".\libusb"
-  - set PATH=%PATH%;%APPVEYOR_BUILD_FOLDER%\libusb\MS32\dll
+  - ps: Copy-Item ".\libusb\$env:LIBUSB_DLL_FOLDER\dll\*.dll" "." -Force -verbose
 
 build_script:
-  - pip install -e .
-  - ykman -v
-
-test_script:
-  - python setup.py test
```
emlun: Why was the test step removed?

dagheyman:
I tried to address most of the comments. Both appveyor and travis now has conditional deploys, so we can test as many python versions as we want but only build artifacts on some of them.

PR #157, opened 2018-10-11
Add class methods for inspecting verify fail (0x63cX) status words

emlun:
An alternative approach would be to add similar methods to APDUError instead.

### DIFF
--- a/ykman/piv.py
+++ b/ykman/piv.py
@@ -256,12 +256,12 @@ class SW(IntEnum):
     AUTHENTICATION_BLOCKED = 0x6983
     INCORRECT_PARAMETERS = 0x6a80
 
-    @classmethod
-    def is_verify_fail(cls, sw):
+    @staticmethod
+    def is_verify_fail(sw):
         return 0x63c0 <= sw <= 0x63c3
 
-    @classmethod
-    def tries_left(cls, sw):
+    @staticmethod
+    def tries_left(sw):
         # Blocked, 0 tries left.
         if sw == SW.AUTHENTICATION_BLOCKED:
             return 0
--- a/ykman/piv.py
+++ b/ykman/piv.py
@@ -256,6 +256,16 @@ class SW(IntEnum):
     AUTHENTICATION_BLOCKED = 0x6983
     INCORRECT_PARAMETERS = 0x6a80
 
+    @classmethod
+    def is_verify_fail(cls, sw):
+        return 0x63c0 <= sw <= 0x63c3
+
+    @classmethod
+    def tries_left(cls, sw):
+        if not cls.is_verify_fail(sw):
+            raise ValueError('Not a verify fail status word: %x' % sw)
+        return sw - 0x63c0
+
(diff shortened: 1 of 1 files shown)


dagheyman:
I'd suggest making `PivController._parse_tries_left` a public method instead.

emlun:
Agreed offline to move `PivController._parse_tries_left` into `SW` instead.

dainnilsson:
Why are these classmethods that take an instance as an argument, instead of properties?

emlun:
Because the enumeration doesn't have constans for all the possible `0x63cX` values:

- `VERIFY_FAIL_RETRY_0 = 0x63c0`
- `VERIFY_FAIL_RETRY_1 = 0x63c1`
- `VERIFY_FAIL_RETRY_2 = 0x63c2`
- `VERIFY_FAIL_RETRY_3 = 0x63c3`
- `COUNTER_VALUE_4 = 0x63c4`
- ... 
- `COUNTER_VALUE_15 = 0x63cf`

and without that it's not possible to have "instances" of the enumeration on which to invoke the properties. We could of course add those enumeration constants, but then `SW(sw).is_verify_fail` would _still_ crash if given a `sw` that doesn't match any of the constants.

dainnilsson:
Ah. In that case, `@staticmethod`?

emlun:
Ah, I didn't know there's that option. Sure!

ykman/piv.py
```diff
@@ -256,6 +256,25 @@ class SW(IntEnum):
     AUTHENTICATION_BLOCKED = 0x6983
     INCORRECT_PARAMETERS = 0x6a80
 
+    @classmethod
+    def is_verify_fail(cls, sw):
+        return 0x63c0 <= sw <= 0x63c3
```
dainnilsson: It's not quite clear to me why we're just considering values 0-3 as valid here, but the `tries_left` function handles a much larger span.

emlun: Yeah, I agree. The APDU list says

High byte | Low byte | Severity | Explanation
-- | -- | -- | --
63 | C0 | W | Verify fail, no try left.
63 | C1 | W | Verify fail, 1 try left.
63 | C2 | W | Verify fail, 2 tries left.
63 | C3 | W | Verify fail, 3 tries left.
63 | CX | W | The counter has reached the value 'x' (0 = x = 15) (command dependent).

so maybe we should expand `is_verify_fail`? I'm not sure which is more correct.

In actuality, the reason for the discrepancy is that `tries_left` was moved here from `PivController._parse_tries_left` and `is_verify_fail` was newly written before we decided to move `_parse_tries_left`.

emlun: Bump

dagheyman: I think 63CX should be seen as a "Verify fail" as well, no?

dainnilsson: Agreed

emlun: All right, will do!

Issue #160, opened 2018-10-20
encoding error prevents oath key deletion


otsuarez:
Hi, 
I'd been adding oath keys to my yubikey without any issues. But then I write the command on  Notes app (using MacOS High Sierra), which happened to had smart quotes substitution enabled. I didn't realized that at first, so i just write the command, copy and paste on the cli.
Got the error then trying to get the code:
```sh
mbp:~$ ykman oath code github
github:username@acme.com    [Touch Credential]
Error: 'ascii' codec can't encode character u'\u201c' in position 0: ordinal not in range(128)
mbp:~$
```
Because of the smart quotes substitution, the `"` quote was replaced with `“`:
```
mbp:~$ ykman oath list
“github:username@acme.com”
```
I also tried,  unsuccessfully, to delete the key:
```sh
mbp:~$ ykman oath delete “github:username@acme.com”
Error: 'ascii' codec can't encode character u'\u201c' in position 0: ordinal not in range(128)
mbp:~$ ykman list
YubiKey 4 [FIDO+CCID] Serial: 1234567
mbp:~$
```
Not a major issue, but worth the issue report. 

How can I delete that entry?

Last not least, great tool!

dagheyman:
Thanks for the report!

What version of ykman are you running, and how did you install it?
Your example credential seems to work fine for me:
```
$ ykman oath add “github:username@acme.com” abba
$ ykman oath list
“github:username@acme.com”
$ ykman oath delete “github:username@acme.com”
Delete credential: “github:username@acme.com” ? [y/N]: y
Deleted “github:username@acme.com”.
```

Could you also run the command with `--log-level debug` to get more logs ?

otsuarez:
More info:
ykman was installed via brew

brew install ykman

version:

mbp:~$ ykman --version
YubiKey Manager (ykman) version: 0.7.1
Libraries:
    libykpers 1.19.0
    libusb 1.0.22
mbp:~$

Using --log-level:

mbp:~$ ykman  --log-level debug oath  delete  “github:username@acme.com”
2018-10-22T07:13:06-0400 INFO [ykman.logging_setup.setup:59]
Initialized logging for ykman version: 0.7.1
2018-10-22T07:13:06-0400 DEBUG
[ykman.descriptor.Descriptor.open_device:75] transports: 0x4,
self.mode.transports: 0x6
2018-10-22T07:13:06-0400 DEBUG [ykman.descriptor.open_device:80]
[...]

PR #162, opened 2018-10-23
oath: use explicit unicode string when needed

dagheyman:
See #160 
For Python 2 compatibility.


dagheyman:
> Maybe add a CLI test (or change an existing one) that exposes the bug?
Done!

PR #164, opened 2018-10-23
WIP: Postpone opening device until needed.

dainnilsson:
This can probably be cleaned up quite a bit, and needs testing, but might be a reasonable approach.

### DIFF
--- a/ykman/cli/__main__.py
+++ b/ykman/cli/__main__.py
@@ -34,7 +34,7 @@ from ..native.pyusb import get_usb_backend_version
 from ..driver_otp import libversion as ykpers_version
 from ..descriptor import (get_descriptors, list_devices, open_device,
                           FailedOpeningDeviceException)
-from .util import click_skip_on_help, UpperCaseChoice
+from .util import UpperCaseChoice
 from .info import info
 from .mode import mode
 from .otp import otp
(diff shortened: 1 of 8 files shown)


ykman/cli/util.py
```diff
@@ -67,14 +67,11 @@ def inner(ctx, param, val):
     return wrap
 
 
-def click_skip_on_help(f):
+def click_postpone_exection(f):
```
emlun: Missing a "u": `click_postpone_execution`

ykman/cli/util.py
```diff
@@ -67,14 +67,11 @@ def inner(ctx, param, val):
...
     @functools.wraps(f)
     def inner(*args, **kwargs):
         ctx = click.get_current_context()
-        for arg in ctx.help_option_names:
-            if arg in sys.argv:
-                return
-        f(*args, **kwargs)
+        ctx.obj['dev'].on_resolves.append(lambda: f(*args, **kwargs))
```
emlun: This should probably be encapsulated with an `on_resolve` method or the like instead. Then that could also call the argument immediately if it's already resolved.

(I kind of want to call it `.then(lambda...)` like Promises in JS, but that's probably not a very good idea...)

ykman/cli/fido.py
```diff
@@ -74,6 +75,7 @@ def info(ctx):
     """
     Display status of FIDO2 application.
     """
+    ctx.obj['dev'].get()
```
emlun: This looks like a side effect free expression, which therefore looks useless since it's not assigned to anything. Maybe rename the `get` method to something like `resolve`, or add an alias?

ykman/cli/fido.py
```diff
@@ -74,6 +75,7 @@ def info(ctx):
...
     controller = ctx.obj['controller']
```
emlun: Related to the previous point - the relationship between this and the previous line isn't obvious. Maybe we could express this as a method on the `_DeviceResolver` class instead?

Maybe `ctx.obj['dev'].set_controller(...)` instead of `ctx.obj['controller'] = ...`?

Or maybe `ctx.obj['dev'].get_fido_controller()`, `.get_otp_controller()` etc., which call `.get()` internally?

ykman/cli/__main__.py
```diff
@@ -119,6 +119,37 @@ def _run_cmd_for_single(ctx, cmd, transports):
...
+                )
+            else:
+                self._dev = _run_cmd_for_single(
+                    self.ctx,
+                    self.subcmd,
+                    self.transports
+                )
+            self.ctx.call_on_close(self._dev.close)
+            for f in self.on_resolves:
+                f()
```
emlun: Maybe pass `self._dev` as an argument? ...might be a bit too much magic.

emlun:
>I don't think any of your comments apply to the newer code, am I right?

Probably not, I'll re-review as soon as I can.

dainnilsson:
Squash merged onto master.

PR #172, opened 2018-11-07
Merge SW enums

emlun:
This would merge into #169.


emlun:
Right now this leaves `piv.SW` as a standalone enum with some overlap with `driver_ccid`; perhaps we want to merge that into `driver_ccid.SW` as well.

dagheyman:
Yeah, maybe we can get away with a single `SW` enum in `driver_ccid.py`.

PR #181, opened 2018-11-21
Extract method PivController.supported_algorithms

emlun:
This allows us to tailor the GUI optimally for each kind of YubiKey (i.e., not show options for unsupported algorithms).


ykman/cli/piv.py
```diff
@@ -223,22 +224,25 @@ def generate_key(
 
     _ensure_authenticated(ctx, controller, pin, management_key)
 
-    algorithm = ALGO.from_string(algorithm)
+    algorithm_id = ALGO.from_string(algorithm)
```
dagheyman: Why this name change?

emlun: Because previously, `algorithm` changed type halfway so the error message would show "Algorithm 6 is not supported on this YubiKey".

PR #187, opened 2018-12-11
fido: handle too long PINs better

dagheyman:


dagheyman:
ready for re-review. same spirit as https://github.com/Yubico/yubikey-manager-qt/pull/148

PR #189, opened 2018-12-17
piv: trim metadata in beginning of pem certs

dagheyman:


dagheyman:
Turned out the actual trimming was not needed, just relaxing the PEM check (allow noise before certificate starts) was enough to load the certs. This seems to work with certificate chains as well, all though that needs more testing.

PR #200, opened 2019-01-02
Fix FASC-N encoding

solidgoldbomb:
This is an equivalent change to the one that merged in yubico-piv-tool in https://github.com/Yubico/yubico-piv-tool/pull/178.

Tested on yubikey neo.

Correctly decodes as ```[9999-9999-999999-0-1-0000000000300001]``` on this [online decoder site](http://www.christhilges.com/fasc-n/index.php).

### DIFF
--- a/ykman/piv.py
+++ b/ykman/piv.py
@@ -999,12 +999,18 @@ class PivController(object):
         return certs
 
     def update_chuid(self):
+        # Non-Federal Issuer FASC-N
+        # [9999-9999-999999-0-1-0000000000300001]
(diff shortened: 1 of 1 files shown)


dagheyman:
Thanks!

solidgoldbomb:
Thanks for merging.  Are there any other tools that need a similar change to line them all up?  ykman and yubico-piv-tool are the only ones I'm aware of.

dagheyman:
@solidgoldbomb No these two would be the ones. There is YubiKey PIV Manager, but that is being phased out and no new releases are planned. There is also the YubiKey Minidriver for Windows, but that is not on GitHub. I have reached out to the maintainers of that to look over this.

Thanks again for the find and patches!

Issue #183, opened 2018-11-30
RFE: Need equivalent to yubico-piv-tool -a write-object


opoplawski:
Trying to do:
```
# yubico-piv-tool -a write-object --id 0x5FC10C
Failed authentication with the application.
```
Which is apparently because yubico-piv-tool does not support the PIN derived management key on our YubiKey 4s.  ykman appears to be the successor to yubico-piv-tool, but it does not appear to have this functionality.

dagheyman:
Thanks, we'll look into adding this functionality in future versions of ykman.

In the meantime I'd recommend [contacting Yubico Support](https://support.yubico.com/support/tickets/new) for guidance on possible workarounds.

opoplawski:
Will do.  Thanks.

smrt28:
Could you please explain what does "PIN derived management key" mean? As far as I understand, you derive mgm-key from the password by hashing in some (mostly UI based) apps to make the things more user-friendly since the user doesn't have to remember "another mgm-key".

opoplawski:
> In the meantime I'd recommend [contacting Yubico Support](https://support.yubico.com/support/tickets/new) for guidance on possible workarounds.

FWIW - Yubico Support didn't have any suggestions.

dagheyman:
@smrt28 Yes, this is what YubiKey PIV Manager did, described here: https://developers.yubico.com/yubikey-piv-manager/PIN_and_Management_Key.html

In yubikey-manager however, we instead generate a random management key and store it on the device, in a slot protected by the PIN. See the command `$ ykman piv change-management-key --protect` 

The tool do support using devices setup with the old derivation scheme as well, to be compatible.

smrt28:
@dagheyman "slot protected by PIN" is something I'm looking for. I want to store an arbitrary secret on Yubikey which would be protected by the PIN and use it as a key-file for LUKS disk encryption. Could you please provide me with some details about how I can reach that? Thanks

dagheyman:
@smrt28 Look into https://web.archive.org/web/20170921144353/http://nvlpubs.nist.gov:80/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf and https://github.com/Yubico/yubikey-manager/blob/master/ykman/piv.py . We are using the slot for printed information to store the management key.

smrt28:
[...]

PR #208, opened 2019-01-21
piv: add read-object and write-object commands

dagheyman:
Add commands for writing and reading arbitrary PIV objects. The read commands should prompt for the pin if needed the write command should prompt for pin or management key if needed.

requested in #183

### DIFF
--- a/test/on_yubikey/cli_piv/test_misc.py
+++ b/test/on_yubikey/cli_piv/test_misc.py
@@ -1,6 +1,5 @@
 from ..util import ykman_cli
-from .util import PivTestCase
-
+from .util import PivTestCase, DEFAULT_MANAGEMENT_KEY
 
 class Misc(PivTestCase):
 
@@ -11,3 +10,11 @@ class Misc(PivTestCase):
     def test_reset(self):
         output = ykman_cli('piv', 'reset', '-f')
         self.assertIn('Success!', output)
+
+    def test_write_read_object(self):
+        ykman_cli(
+            'piv', 'write-object', '--id',
+            '0x5f0001', '-m', DEFAULT_MANAGEMENT_KEY,
+            '-', input='test data')
+        output = ykman_cli('piv', 'read-object', '--id', '0x5f0001')
+        self.assertIn('test data', output)
--- a/test/on_yubikey/cli_piv/test_misc.py
+++ b/test/on_yubikey/cli_piv/test_misc.py
@@ -13,8 +13,8 @@ class Misc(PivTestCase):
 
     def test_write_read_object(self):
         ykman_cli(
-            'piv', 'write-object', '--id',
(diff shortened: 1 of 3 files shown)


test/on_yubikey/cli_piv/test_misc.py
```diff
@@ -11,3 +10,11 @@ def test_info(self):
...
         output = ykman_cli('piv', 'reset', '-f')
         self.assertIn('Success!', output)
+
+    def test_write_read_object(self):
+        ykman_cli(
+            'piv', 'write-object', '--id',
+            '0x5f0001', '-m', DEFAULT_MANAGEMENT_KEY,
+            '-', input='test data')
+        output = ykman_cli('piv', 'read-object', '--id', '0x5f0001')
+        self.assertIn('test data', output)
```
emlun: Why not `assertEqual[s]`?

ykman/cli/piv.py
```diff
@@ -792,6 +792,68 @@ def unblock_pin(ctx, puk, new_pin):
             show_default=False, hide_input=True, err=True)
     controller.unblock_pin(puk, new_pin)
 
+@piv.command('read-object')
+@click_pin_option
+@click.pass_context
+@click.option(
+        '-i', '--id', 'object_id',
+        callback=lambda ctx, param, value: int(value, 16),
+        metavar='HEX', help='Id of PIV object.', required=True)
```
emlun: Why not make `--id` a required positional parameter instead?

ykman/cli/piv.py
```diff
@@ -792,6 +792,68 @@ def unblock_pin(ctx, puk, new_pin):
...
+
+@piv.command('write-object')
+@click_pin_option
+@click_management_key_option
+@click.pass_context
+@click.option(
+        '-i', '--id', 'object_id',
+        callback=lambda ctx, param, value: int(value, 16),
+        metavar='HEX', help='Id of PIV object.', required=True)
+@click.argument('data', type=click.File('rb'), metavar='DATA')
```
emlun: It isn't clear from the help text that this should be a file name, or that `-` resolves to standard input.

ykman/cli/piv.py
```diff
@@ -792,6 +792,68 @@ def unblock_pin(ctx, puk, new_pin):
...
+    do_read_object()
+
+@piv.command('write-object')
+@click_pin_option
+@click_management_key_option
+@click.pass_context
+@click.option(
+        '-i', '--id', 'object_id',
+        callback=lambda ctx, param, value: int(value, 16),
+        metavar='HEX', help='Id of PIV object.', required=True)
```
emlun: Why not make `--id` a required positional parameter instead?

dagheyman:
Updated PR to address all comments. Ready for re-review!

PR #213, opened 2019-01-28
parse_certs: fail on empty list

dagheyman:
`ykman.util.parse_certificates` returned an empty list on pem files without certificates (such as private keys). This leads to problems in yubikey-manager-qt, where this would be interpreted as the data containing certificates.

### DIFF
--- a/ykman/util.py
+++ b/ykman/util.py
@@ -473,7 +473,9 @@ def parse_certificates(data, password):
                         PEM_IDENTIFIER + cert, default_backend()))
             except Exception:
                 pass
-        return certs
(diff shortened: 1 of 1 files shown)


dagheyman:
Thea idea is that it should throw an error if no valid certs are found. It's being used for heuristics here: https://github.com/Yubico/yubikey-manager-qt/blob/master/ykman-gui/py/yubikey.py#L567

emlun:
Oh, I missed that execution continues from there - I thought it would return `None` in that case. Yes, this makes sense then.

PR #215, opened 2019-01-29
Fix how the context is resolved

rgerganov:
On Python3 dict.items() returns a view object and changing dictionary
values while iterating this view may lead to unexpected results when we
resolve the context like an attempt to resolve the same object twice.

This patch fixes this by creating a copy of the underlying dict in the
resolve() method.

### DIFF
--- a/ykman/cli/util.py
+++ b/ykman/cli/util.py
@@ -81,7 +81,7 @@ class YkmanContextObject(MutableMapping):
     def resolve(self):
         if not self._resolved:
             self._resolved = True
-            for k, f in self._objects.items():
+            for k, f in self._objects.copy().items():
                 self._objects[k] = f()
 
     def __getitem__(self, key):


dagheyman:
LGTM, @dainnilsson opinions?

dainnilsson:
It's undesirable to modify the collection as its being iterated over, but I think a cleaner solution would be to instead create a new dict with the resolved values to replace the one with functions, than to copy the one with functions and modify the original dict in place. I would suggest replacing the for loop with:

```python
self._objects = {(k, f()) for k, f in self._objects.items()}
```

rgerganov:
@dainnilsson well, it turns out it's not so simple because resolving the second element of the `OrderedDict` requires the first one to be resolved. This can't work if we construct a temporary dict and replace the original one at the end.

I can think of only 2 options here:
1. go with my original patch
2. rewrite `YkmanContextObject` to track both "resolved" and "unresolved" objects explicitly

Let me know which one you prefer (or you come up with something else) and I will update

dainnilsson:
Ah, you're right. I forgot about the ordering requirement. In that case, the first patch is probably the easiest thing to do. Sorry for the extra noise!

dagheyman:
Merging original patch, thanks!

PR #220, opened 2019-02-19
static pw: fix support for DE layout

dagheyman:


dagheyman:
This pull request **introduces 1 alert** when merging f7d44cc005cad06bd200e4fbbab481a853370f00 into a79f8f16da14b4ef39bcc057df8fb652a0b27348 - [view on LGTM.com](https://lgtm.com/projects/g/Yubico/yubikey-manager/rev/pr-8cf6df497db998e4158c06f84ab32a28a7294af5)

**new alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

ykman/util.py
```diff
@@ -351,14 +352,11 @@ def modhex_encode(value):
...
+        keyboard_layout=KEYBOARD_LAYOUT.MODHEX,
         blacklist=DEFAULT_PW_CHAR_BLACKLIST):
-    data = os.urandom(length)
-    keys = ''.join([
-        k for k in keyboard_layout.value.keys() if k not in blacklist]).encode()
-    return bytes(
-            bytearray(six.indexbytes(
-                keys, d % len(keys)) for d in six.iterbytes(data)))
+    chars = [k for k in keyboard_layout.value.keys() if k not in blacklist]
+    return ''.join([random.SystemRandom().choice(chars) for _ in range(length)])
```
dainnilsson: Any reason for not just using https://docs.python.org/2/library/random.html#random.choice here?

dagheyman: Just this message,
```
The pseudo-random generators of this module should not be used for security purposes. Use os.urandom() or SystemRandom if you require a cryptographically secure pseudo-random number generator.
``` 
happy to change if the non-system one is enough.

dainnilsson: Ah, I see. I'd probably use SystemRandom then, but move the instantiation outside of the loop so it's only called once. I also noted that it's `Not available on all systems.`, which might be a problem?

dagheyman: Good point, will do that. The not available part is probably just because it uses `os.urandom`(see https://docs.python.org/3/library/os.html#os.urandom), which was used before, so should be fine.

PR #221, opened 2019-02-20
add support fort listing smart card readers

dagheyman:
- add `-r/--reader` flag to ykman list, for listing available smart card readers (both yubikeys and external)
- relax the filtering, so that it's possible to select reader with any substring of the reader name. For example `OMNI` would now work on a `HID Global OMNIKEY 5022 Smart Card Reader` reader, before `HID` was required.

### DIFF
--- a/ykman/cli/__main__.py
+++ b/ykman/cli/__main__.py
@@ -32,7 +32,7 @@ from ykman import __version__
 from ..util import TRANSPORT, Cve201715361VulnerableError, YUBIKEY
 from ..native.pyusb import get_usb_backend_version
 from ..driver_otp import libversion as ykpers_version
-from ..driver_ccid import open_devices as open_ccid
+from ..driver_ccid import open_devices as open_ccid, list_readers
 from ..device import YubiKey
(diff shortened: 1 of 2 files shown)


ykman/cli/__main__.py
```diff
@@ -194,11 +194,17 @@ def resolve_device():
...
 @click.pass_context
-def list_keys(ctx, serials):
+def list_keys(ctx, serials, readers):
     """
     List connected YubiKeys.
     """
+    if readers:
+        for reader in list_readers():
+            click.echo(reader.name)
+        ctx.exit()
```
emlun: Just a thought: why not print both readers and YubiKeys?

dagheyman: Well YubiKeys will show up as well, if the are in CCID mode. I see this as a kind of special case where only reader-names are interesting (NFC or RDP use case), and listing devices again based on descriptors adds noise.

emlun: Ok :+1:

PR #235, opened 2019-04-25
building centos7 compatible .rpm packages in CI

dagheyman:
this PR provides a docker image to build centos7 compatible .rpm packages.

Current limitations:
- hard-coded version of ykman in .spec file, gets a release from dev.y.c

Installation instructions:

```
yum install epel-release
yum install --enablerepo=epel-testing python3-yubikey-manager.rpm
yum install yubikey-manager.rpm
sudo systemctl start pcscd
```


docker/centos7-rpm/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM centos:7
+ENV LANG en_US.UTF-8
+ENV LC_ALL en_US.UTF-8
+RUN yum -y install gcc rpm-build rpm-devel rpmlint make python bash coreutils diffutils patch rpmdevtools wget python-setuptools
+RUN yum -y install epel-release
```
emlun: Looks like this can be on the same line as the rest?

docker/centos7-rpm/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN yum-config-manager --enable epel-testing
+RUN yum repolist
```
emlun: Looks like this is unnecessary?

dagheyman:
good points, simplified the Dockerfile a bit

PR #237, opened 2019-04-30
otp: use 64 byte challenges for TOTP calculation

dagheyman:


ykman/otp.py
```diff
@@ -190,6 +190,8 @@ def calculate(
                 challenge = time_challenge(time.time())
             else:
                 challenge = time_challenge(challenge)
+            # Always use 64 byte challenges for TOTP, assume HMAC_LT64
+            challenge = challenge.ljust(64, '0')
```
dainnilsson: This can't simply be padded with nulls, it must be padded with a character that is different from the last byte of the challenge.

dainnilsson: Suggestion:

`challenge.ljust(64, b'\1' if challenge.endswith(b'\0') else b'\0')`

dagheyman: Thanks, updated with your suggestion.

PR #241, opened 2019-05-13
Add -e Initial time since the UNIX epoch from which to calculate the counter value.

xqin:
Initial time since the UNIX epoch from which to calculate the counter value.
Defaults to 0 (no offset).

example:

```
ykman oath code -e 123456 -s hello
```

### DIFF
--- a/ykman/cli/oath.py
+++ b/ykman/cli/oath.py
@@ -332,7 +332,9 @@ def list(ctx, show_hidden, oath_type, period):
 @click.argument('query', required=False, default='')
 @click.option('-s', '--single', is_flag=True, help='Ensure only a single '
               'match, and output only the code.')
-def code(ctx, show_hidden, query, single):
+@click.option('-e', '--epoch', type=click.INT, default=0, help='Initial time since the UNIX epoch from '
+              'which to calculate the counter value. Defaults to 0 (no offset).')
+def code(ctx, show_hidden, query, single, epoch):
     """
     Generate codes.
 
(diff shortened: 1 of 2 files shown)


dagheyman:
Hi, thanks for the contribution!

Before we review the code in detail, could you clarify a bit what use case you are solving with this?

dagheyman:
> And I have encountered the situation needs to be specified T0 in the real system.

Could  you describe this a bit more? What kind of systems are these?

xqin:
The internal VPN network uses TOTP as the password, but the TOTP is generated with the specified T0, that's all.  @dagheyman

dagheyman:
@xqin Thanks, interesting. What TOTP client do you typically use that let's you specify this? Is it a custom/in-house built thing? Would it be sufficient to specify the timestamp itself in a command line flag as opposed to the offset?

Sorry for all the questions, but I have not seen this before and I'm trying to figure out how rare/common this scenario is.

xqin:
https://github.com/speakeasyjs/speakeasy#totpoptions--string
https://github.com/xqin/hotp-cli

Currently I have created a hotp-cli command line tool using this module of nodejs.

But using this method requires passing the TOTP token as a parameter to the command line.

I want to use a safer way to save the TOTP token to Yubikey, and then use the ykman command to get the result, so I have the PR now.

Sorry for my english, The above content was translated by Google Translate, I hope you can understand what I want to explain.

xqin:
Using a private T0 can make the TOTP Token more secure, even if the Token is compromised, the other party will not be able to generate the correct verification code.

For example, I lost Yubikey, someone picked it up, because he doesn't know the value of T0, then it is always wrong when he tries to use the verification code calculated by the standard time.

dagheyman:
Sorry, I'm not going to merge this, the use case is to narrow/rare. I recommend setting a password to protect the credentials in the case of a lost device.

Thanks anyway!

PR #247, opened 2019-05-22
ykman list: list based on descriptors

dagheyman:
- Add a `name` property on the descriptor object (uses PID and version to come up with a name)
- Simplify SKY specific code in list command
- If any device failed to open, list it based on the information available in the descriptor

Examples with FIDO only device on Win 10:
```
$ ykman.exe list
Security Key NFC [FIDO]

$ ykman.exe info
Error: Failed connecting to Security Key NFC [FIDO]
```

### DIFF
--- a/ykman/descriptor.py
+++ b/ykman/descriptor.py
@@ -27,7 +27,7 @@
 
 from __future__ import absolute_import
 
-from .util import PID, TRANSPORT, Mode
+from .util import PID, TRANSPORT, Mode, YUBIKEY
 from .device import YubiKey
 from .driver_ccid import open_devices as open_ccid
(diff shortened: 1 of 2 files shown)


ykman/descriptor.py
```diff
@@ -70,6 +70,15 @@ def mode(self):
     def key_type(self):
         return self._key_type
 
+    @property
+    def name(self):
+        if self.key_type == YUBIKEY.SKY and self.version <= (5, 0, 0):
```
emlun: Shouldn't this be `<` for the U2F key?

dagheyman: That's correct, good catch!

PR #248, opened 2019-05-22
Refactor: Extract nested function print_device in list_keys

emlun:


ykman/cli/__main__.py
```diff
@@ -201,6 +202,18 @@ def list_keys(ctx, serials, readers):
     """
     List connected YubiKeys.
     """
+
+    def print_device(dev, serial):
```
dagheyman: Maybe call it `_print_device` ?

emlun: Sure, but is that necessary for a nested function?

emlun: Done.

PR #250, opened 2019-06-07
add support for new preview device

dagheyman:


ykman/device.py
```diff
@@ -284,7 +285,8 @@ def __init__(self, descriptor, driver):
             if config.nfc_supported:
                 self.device_name = 'Security Key NFC'
         elif self._key_type == YUBIKEY.YK4:
-            if (5, 1, 0) > self.version >= (5, 0, 0):
+            if (5, 1, 0) > self.version >= (5, 0, 0) or \
```
emlun: Although not introduced by this PR, I think the inequalities look a little confusing in this direction.

```suggestion
            if (5, 0, 0) <= self.version < (5, 1, 0) or \
```

PR #252, opened 2019-06-13
draft: extra openpgp features

dagheyman:

PR #254, opened 2019-06-14
Fix flake8 complaints

emlun:


emlun:
Note: this has some conflicts with #252

PR #257, opened 2019-06-17
Extract utility classes for choices of enum members

emlun:
This replaces most uses of `UpperCaseChoice` and post-processing with uses of the new types `EnumNameChoice` and `EnumValueChoice`.

### DIFF
--- a/ykman/cli/util.py
+++ b/ykman/cli/util.py
@@ -54,6 +54,22 @@ class UpperCaseChoice(click.Choice):
                 value, ', '.join(self.choices)), param, ctx)
 
 
+class EnumChoice(UpperCaseChoice):
+    """
+    Use an enum as the definition for a choice option.
+
+    Enum member names MUST be all uppercase. Options are not case sensitive.
+    Underscores in enum names are translated to dashes in the option choice.
+    """
+    def __init__(self, choices_enum):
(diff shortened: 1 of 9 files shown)


ykman/cli/util.py
```diff
@@ -51,6 +54,38 @@ def convert(self, value, param, ctx):
...
+    Underscores in enum names are translated to dashes in the option choice.
+    """
+    def __init__(self, choices_enum):
+        super().__init__([v.name.replace('_', '-') for v in choices_enum])
+        self.choices_enum = choices_enum
+
+    def convert(self, value, param, ctx):
+        name = super().convert(value.replace('-', '_'), param, ctx)
+        return self.choices_enum[name]
+
```
dagheyman: Maybe we could refactor KEY_SLOT and only have one EnumChoice instead

dagheyman:
LGTM, although it does feel like the docstring in `EnumChoice` is screaming at me `"Enum member names MUST be all uppercase."`.. Your call if you want to soften the capitalization :)

emlun:
Hah, yeah, that's a habit from working on specs where those words have strictly defined meanings per [RFC 2119](https://tools.ietf.org/html/rfc2119). If you don't mind I'll leave it in. :)

PR #258, opened 2019-06-19
Fix some broken tests

emlun:
- `TestExternalLibraries` does not need a YubiKey
- Some tests attempted RSA key generation on ROCA vulnerable devices
- OATH and/or OpenPGP tests sometimes crashed for no apparent reason, seems to be related to `test_mode_command` running just before them.


dagheyman:
Looks good, just a thought, should we do oath/openpgp reset after each test as well? Just to minimize the chance of the previous test impacting the next one.

emlun:
Sure, but let's do that a little later. It would create a lot of conflicts with my other development branch (the internal integration test mini-framework) which is based on this.

PR #260, opened 2019-06-25
Convert skip decorators to internal mini-framework for multiple-YubiKey tests

emlun:
Currently, the integration tests in `test/on_yubikey/` are built under the assumption that only one YubiKey is connected, and use `@unittest.skipIf` decorators to skip tests that don't apply to the currently connected YubiKey. This PR replaces the current system with one where multiple YubiKeys can be connected at once, the tests declare which YubiKeys they are (not) compatible with, and the framework takes care of wiring each test together with how to run it for each compatible YubiKey.

I've done my best to make the code comprehensible, but I'm sure there's a lot that can be improved there. I'm also happy to discuss and polish on the names of the various functions and decorators.

I also haven't figured out what to do about `test_interfaces`, since that test is unique in exercising all transports and also by going directly through each driver to get a device connection. I think it's no longer guaranteed that the test uses the same YubiKey throughout the test. Maybe we should just let that remain a special case and skip it unless exactly one YubiKey is connected?

Note that the changes to the tests themselves are mostly whitespace, this isn't nearly as invasive as the diff makes it look.

Looking forward to your thoughts, ideas and criticisms! :smile:

## Using it

 1. Remove `ykman_cli` and `open_device` imports from the test module.
 2. Wrap the module's test classes in a `def additional_tests(arg)` function. Either decorate it with `@device_test_suite(transports)` and name the argument `open_device`, or decorate it with `@cli_test_suite` and name the argument `ykman_cli`.
 3. Make `additional_tests` return a list containing all the test classes.
 4. Write tests as usual, as if there was only one YubiKey connected.
 5. Use the decorators from `test.on_yubikey.framework.yubikey_conditions` to indicate which YubiKeys each test is applicable to. By default, with no decorators, the test is assumed to be applicable to all YubiKeys.
 6. Run the test suite with the `DESTRUCTIVE_YUBIKEY_TEST_SERIALS` environment variable set to a comma-separated list of serial numbers of YubiKeys to use for the tests.

## How it works

The system utilizes the [`additional_tests()`][addt] hook of setuptools's test discovery, and conceptually transforms this:

```python
@device_test_suite(TRANSPORT.CCID)
def additional_tests(open_device):

    class PivKeyManagement(unittest.TestCase):

        def setUp(self):
            self.dev = open_device()
            self.controller = PivController(self.dev.driver)

        def tearDown(self):
            self.dev.driver.close()

        def generate_key(self, slot, alg=ALGO.ECCP256, pin_policy=None):
            self.controller.authenticate(DEFAULT_MANAGEMENT_KEY)
            public_key = self.controller.generate_key(
                slot, alg, pin_policy=pin_policy,
                touch_policy=TOUCH_POLICY.NEVER)
            self.reconnect()
            return public_key

        @yubikey_conditions.supports_piv_touch_policies
        def test_delete_certificate_requires_authentication(self):
            self.generate_key(SLOT.AUTHENTICATION)

            with self.assertRaises(APDUError):
                self.controller.delete_certificate(SLOT.AUTHENTICATION)

            self.controller.authenticate(DEFAULT_MANAGEMENT_KEY)
            self.controller.delete_certificate(SLOT.AUTHENTICATION)

        @yubikey_conditions.is_fips
        def test_pin_policy_never_blocked_on_fips(self):
            with self.assertRaises(APDUError):
                self.generate_key(pin_policy=PIN_POLICY.NEVER)

    return [PivKeyManagement]
```

into this:

```python
class PivKeyManagement_CCID_5.0.2_7652135(unittest.TestCase):

    def setUp(self):
        self.dev = open_device(transports=TRANSPORT.CCID, serial=7652135)
        self.controller = PivController(self.dev.driver)

    def tearDown(self):
        self.dev.driver.close()

    def generate_key(self, slot, alg=ALGO.ECCP256, pin_policy=None):
        self.controller.authenticate(DEFAULT_MANAGEMENT_KEY)
        public_key = self.controller.generate_key(
            slot, alg, pin_policy=pin_policy,
            touch_policy=TOUCH_POLICY.NEVER)
        self.reconnect()
        return public_key

    def test_delete_certificate_requires_authentication(self):
        self.generate_key(SLOT.AUTHENTICATION)

        with self.assertRaises(APDUError):
            self.controller.delete_certificate(SLOT.AUTHENTICATION)

        self.controller.authenticate(DEFAULT_MANAGEMENT_KEY)
        self.controller.delete_certificate(SLOT.AUTHENTICATION)


class PivKeyManagement_CCID_3.5.0_6513273(unittest.TestCase):

    def setUp(self):
        self.dev = open_device(transports=TRANSPORT.CCID, serial=6513273)
        self.controller = PivController(self.dev.driver)

    def tearDown(self):
        self.dev.driver.close()

    def generate_key(self, slot, alg=ALGO.ECCP256, pin_policy=None):
        self.controller.authenticate(DEFAULT_MANAGEMENT_KEY)
        public_key = self.controller.generate_key(
            slot, alg, pin_policy=pin_policy,
            touch_policy=TOUCH_POLICY.NEVER)
        self.reconnect()
        return public_key


class PivKeyManagement(unittest.TestCase):
    def test_pin_policy_never_blocked_on_fips(self):
        self.skipTest('No YubiKey available for test.')
```

So in this case the test suite is run with two YubiKeys: one of firmware version 3.5.0 and one 5.0.2. Note that

- Each YubiKey gets its own variant of the test class, where the class name contains the firmware version and serial number of the YubiKey and the transport used to communicate with the YubiKey.
- Each variant of the test class gets its own arguments to `open_device`
- One test is missing from the 3.5.0 test class, because the method was decorated with `@yubikey_conditions.supports_piv_touch_policies` which is a condition the YubiKey NEO does not satisfy.
- One test was not satisfied by either YubiKey because it was decorated with `@yubikey_conditions.is_fips`, so it remains in the original test class which now marks all its tests as skipped.

This isn't actually how it's implemented under the hood, but it's the mental model for how the design is intended to work. The actual implementation uses a bunch of Python metaprogramming magic. :smile:

In order to not break how the `setUp` and `tearDown` hooks work, we need to transform each original class definition into multiple variants of the class, one for each YubiKey and transport. That means we can't do this with a simple class decorator, which is why I've used the `additional_tests()` hook. Functions can define and return classes, which is what actually happens inside the `@device_test_suite` decorator. A real `additional_tests()` function should take no parameters and return a `unittest.TestSuite` instance; the `@device_test_suite` decorator expects the decorated function to take one parameter and return a list of `unittest.TestCase` subclasses. `@device_test_suite` then creates a specialized `open_device` function for each YubiKey and transport, and calls the decorated function repeatedly once for each function specialization. Each call to the decorated `additional_tests(open_device)` creates a new set of test classes; the `@device_test_suite` decorator removes any unsupported test methods from each of them and finally assembles and returns a `TestSuite` instance to setuptools's test discovery mechanism.

The `@cli_test_suite` decorator works in much the same way, except it doesn't take a `transports` parameter and it instead provides the decorated `additional_tests(ykman_cli)` with a specialized function for invoking the CLI.

The condition decorators work by attaching a new attribute `_yubikey_conditions` to the test method. The attribute is a set of predicate functions, and each condition decorator adds a predicate to the set. The `@*_test_suite` decorators call each of the predicates in the set, passing as an argument the `ykman.device.YubiKey` handle for the YubiKey for the test method's parent test class, and deletes the method if any predicate returns false. This means that each test method can be decorated with multiple conditions, but the condition decorators themselves cannot reference each other (for "not" variants, for example) because the decorator functions themselves are not predicates.

The condition decorators can also decorate test classes, in which case they are forwarded to all test methods.

[addt]: https://setuptools.readthedocs.io/en/latest/setuptools.html#test-build-package-and-run-a-unittest-suite


dagheyman:
Ok I thought about this a bit, tested it out, and I think we can merge this in. Could you summarize some of the how to use it-info in maybe a README in the tests folder or similar? Just so it's not needed to go back to this PR for that.

emlun:
Sure! I'll try to find some time for that this week.

emlun:
I put some documentation in [doc/integration-tests.adoc](https://github.com/Yubico/yubikey-manager/blob/multi-yubikey-tests/doc/integration-tests.adoc) with a reference from development.adoc, does that look good?

I also added a skip decoration to `test_interfaces` so it'll only run when exactly one YubiKey is present, as noted in the top post. That test isn't really meaningful if there's a chance it grabs different YubiKeys, and it looks nontrivial to get the individual drivers' `open_device` functions to filter by serial number.

dagheyman:
Cool, let's merge this then.

dagheyman:
Of course I found some python2-issues after merging it, could you try to run the tests on python2 and fix any crashes?

dagheyman:
Also `$ python setup.py test (without devices)` seems to fail on python3

```
  File "/home/dag/code/yubikey-manager/test/on_yubikey/framework/__init__.py", line 61, in exactly_one_yubikey_present
    return len(_serials_present) == 1
NameError: name '_serials_present' is not defined
```

emlun:
New PR: https://github.com/Yubico/yubikey-manager/pull/276

PR #266, opened 2019-07-11
Prevent use-after-free when querying usb devices

worr:
This reverts commit 1e8b16ca644c382f912051f144253e2e7bf83bc5.

On OpenBSD at least, this causes a use-after-free in libusb, rendering the entire tool useless. This ends up calling `libusb_init`, which sets up a lot of global context. I'm guessing that that's the root cause, but I'm honestly not sure.

### DIFF
--- a/ykman/descriptor.py
+++ b/ykman/descriptor.py
@@ -47,12 +47,13 @@ class FailedOpeningDeviceException(Exception):
 
 class Descriptor(object):
 
-    def __init__(self, key_type, mode, version, fingerprint, serial=None):
+    def __init__(self, key_type, mode, version, fingerprint, serial=None, backend=None):
         self._logger = logger.getChild('Descriptor')
         self._version = version
         self._key_type = key_type
         self._mode = mode
         self._fingerprint = fingerprint
+        self._backend = backend
 
     @property
     def fingerprint(self):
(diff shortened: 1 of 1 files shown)


dagheyman:
Thanks for noticing and reporting! It was done as a workaround for an issue on macOS where the usb descriptors where being cached incorrectly at times, but it seems like we'll have to look into it some more.

rickard-von-essen:
If this did solve something on macOS maybe this can be wrapped in an OS condition. At least as a temporary workaround.

worr:
I'm not sure that this should be wrapped in a OS guard without testing it on macOS with valgrind or similar first. The UAF only caused a segfault because I have junking turned on in my `malloc.conf`, it may have gone unnoticed otherwise.

worr:
There ends up being a much simpler solution that should hopefully alleviate the issues both on OpenBSD and macOS.

The root cause of the UAF is that pyusb will [call `libusb_exit` sometime after the backend falls out of scope](https://github.com/pyusb/pyusb/blob/v1.0.2/usb/backend/libusb1.py#L712). There's no reference to a backend stored when you query for a device, so a device can live longer than the backend that reported it.

The safest thing to do in this case would be to store a reference to the backend that we used to discover the device. I've changed my code to do that, and verified that this clears up the issue on OpenBSD.

dagheyman:
Thanks for the investigation and contribution! This looks like something we could merge, I will just need to find some time to test it out a bit on different platforms.

dagheyman:
Merged, thanks.

PR #270, opened 2019-07-20
setup.py: Correct dependency on fido2

nbraud:
`yubikey-manager` depends on APIs introduced in `fido2` v0.7.0.
If `fido2` follows semantic versioning conventions, this ought to be `fido2 ~= 0.7.0`.

I ran into this issue while packaging `ykman` v3.0.0.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ import os
 from setuptools import setup
 
 install_requires = [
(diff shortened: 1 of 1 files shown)


dagheyman:
This seems correct, merging. Thanks!

PR #274, opened 2019-08-11
Add the norman keyboard layout as an option for the CLI

bjbishop:
This PR adds support for the [Norman keyboard layout](https://normanlayout.info/index.html) to the ykman CLI.  This layout is extremely rare so very few will ever benefit from this :) but I was mainly seeing if I could accomplish changing the code to make this layout work on my system.

```
  -k, --keyboard-layout [MODHEX|US|DE|NORMAN]
                                  Keyboard layout to use for the static password.  [default: MODHEX]
```
Selecting the `-k NORMAN` now allows me to set and replay a static password correctly using this layout:

```
ykman otp static -k NORMAN -f 2 thisisatesttoseeifthisisgoingtowork
```

Now when mac has the Norman layout selected and I long press the yubikey button, it provides the correct result: `thisisatesttoseeifthisisgoingtowork`


dagheyman:
Hi, thanks for the contribution!

Even if the keyboard layout is rare, it's a great example of how custom layouts can be added to the tool by the community. Will merge it after doing some smoke testing.

dainnilsson:
Might it be better to name this `NORMAN` instead of `NO` to avoid confusing it with Norwegian?

dagheyman:
Good point, I agree. `NORMAN` is a better name for the option.

bjbishop:
Thanks guys!  I'll rename it to `NORMAN` and rebase my commits after a bit

bjbishop:
That's renamed to NORMAN and squashed now

PR #275, opened 2019-08-13
OpenPGP generalization

dainnilsson:
Getting rid of some random number, making things a bit more general and reducing code duplication.


dainnilsson:
@dagheyman @emlun - I've done some various cleanups and made the code more general, in a _mostly_ backwards compatible way. I've also modified the Tlv class to support 2-byte TLVs (marked as such by the first byte having the 5 low bits set) which 

I was also thinking about how a lot of the methods on the controller take the PIN needed as an argument. For example, importing a key will take the admin PIN and verify it, and then import the key itself. Compare this to PIV, where we expose the `authenticate` and `verify` methods, and the caller needs to invoke them separately prior to importing a key.

I think we should change this to use a consistent approach, and I'm leaning more towards how we did it for PIV, as that offers more control to a user of the library (the CLI would remain unchanged of course). Thoughts?

dagheyman:
Looks good to me!

> I think we should change this to use a consistent approach, and I'm leaning more towards how we did it for PIV, as that offers more control to a user of the library (the CLI would remain unchanged of course). Thoughts?

I think you're right in that the approach in PIV offers more control, even though it's nice that the signature of the functions tell you that the PIN is required in the OpenPGP variant.

PR #280, opened 2019-10-11
CI: move CI scipts to GitHub Actions

dagheyman:
The new actions workflows include:
 - A generic build workflow that runs on both python 2 and 3. Runs unit tests, lints with flake8 etc. This workflow is used for the badge in the README.
- macOS and Windows builds (install package, get external deps, run CLI)
- a source package workflow
- a dry-run of the make-ppa script

Dropping:
* PyInstaller based single executable (was not really used for anything)
* Rpm build and yubikey-manager.spec (we are not maintaining that, official one is here https://src.fedoraproject.org/rpms/yubikey-manager/blob/master/f/yubikey-manager.spec)
* docker based debian package build (introducing the make-ppa workflow instead)


.github/workflows/build-ppa.yml
```diff
@@ -0,0 +1,23 @@
+name: Build ppa packages
+# No signing or uploading, just to make sure it doesn't brake
```
emlun:
```suggestion
# No signing or uploading, just to make sure it doesn't break
```

PR #294, opened 2019-12-20
Make sure dlls are loaded properly

fdennis:


fdennis:
> I don't really feel qualified to review this since I don't know DLLs, but the code looks good to me.
> 
> Might want to consider using a context manager for the temporary PATH overrides? But maybe it's too much overhead for too little benefit.

I think a context manager would be a bit too much overhead, as you mention. Also, we have tested this code quite a lot and changes to it would require more testing to make sure the context manager works as it should. It's a nice idea though, perhaps we could add this at a later point if we still want to.

PR #298, opened 2020-01-07
Extract context manager for overriding PATH

emlun:
Maybe a good idea? Maybe too much overhead for too little benefit?

This branches off PR #294.

### DIFF
--- a/ykman/native/libloader.py
+++ b/ykman/native/libloader.py
@@ -42,6 +42,8 @@ import platform
 import ctypes
 import ctypes.util
 
+from ykman.util import override_os_path
+
 
 def _environ_path(name):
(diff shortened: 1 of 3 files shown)


dagheyman:
> too much overhead for too little benefit

I would say so in this case yes.

If we want something like this, it could potentially also live in a new `ykman/native/util.py` file as well. @fdennis What do you think?

fdennis:
I agree with @dagheyman . This would also require us to re-test the dll-loading branch again, which we would like to avoid. If there is something very broken with the current version then yes, we should change it. However, for a small benefit like this I would say it's better to not include this. If we find ourselves needing to set PATHs often, this would be nice though.

PR #300, opened 2020-01-08
cli: hint at permissions when open device fails

dagheyman:
Several OS permission restrictions typically makes ykman fail to open the device, causing a error message. Try to hint at the reason in the message, even if we can't be sure.

This should be triggered by:
- Running a FIDO command on recent Windows 10 without administrator privileges.
- Running a OTP command on macOS Catalina, without making the application to allow "Input Monitoring" in the System Settings. The OS typically prompts for this, at least once.
- Possibly other cases (maybe missing udev on Linux? Not tested.)

Related: https://github.com/Yubico/yubikey-manager-qt/pull/223

### DIFF
--- a/ykman/cli/__main__.py
+++ b/ykman/cli/__main__.py
@@ -100,8 +100,8 @@ def _run_cmd_for_serial(ctx, cmd, transports, serial):
         except FailedOpeningDeviceException:
             ctx.fail(
                 'Failed connecting to a YubiKey with serial: {}. \
-                        Make sure the application have the required permissions.'.format(
-                            serial))
+                        Make sure the application have the required \
+                            permissions.'.format(serial))
 
 
 def _run_cmd_for_single(ctx, cmd, transports, reader=None):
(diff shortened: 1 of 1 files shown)


ykman/cli/__main__.py
```diff
@@ -98,8 +98,10 @@ def _run_cmd_for_serial(ctx, cmd, transports, serial):
                     ctx.fail("Command '{}' is not supported by this device."
                              .format(cmd))
         except FailedOpeningDeviceException:
-            ctx.fail('Failed connecting to a YubiKey with serial: {}'
-                     .format(serial))
+            ctx.fail(
+                'Failed connecting to a YubiKey with serial: {}. \
+                        Make sure you have the correct permissions'.format(
```
emlun: How about "Make sure the application has the necessary permissions" instead? Can we provide any hint to what those permissions are (maybe just a link to some documentation)?

dagheyman: The first suggested change I can do, rephrase it that its the app that may need permissions. But I don't want to add links to online documentation or detect the operating system, since it's mostly guessing anyway.

PR #303, opened 2020-01-15
oath: code -s gave wrong error message

dagheyman:
Found this by accident, but it prints the multiple hits error when there is no hits.

### DIFF
--- a/ykman/cli/oath.py
+++ b/ykman/cli/oath.py
@@ -372,7 +372,7 @@ def code(ctx, show_hidden, query, single):
         _error_multiple_hits(ctx, [cr for cr, c in creds])
 
     elif single and len(creds) == 0:
-        ctx.fail('No credential found.')
+        ctx.fail('No matching credential found.')
 
     if single and creds:
         click.echo(creds[0][1].value)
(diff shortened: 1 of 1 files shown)


dagheyman:
> Shouldn't it still exit with nonzero status code if there is no match?

I'm not sure, can't you see it as the command completed successfully even if there was no match?

emlun:
If you're setting the `-s` option then you presumably want to get exactly one match, so I do think zero matches should be regarded a failure. :slightly_smiling_face:

dagheyman:
> If you're setting the `-s` option then you presumably want to get exactly one match, so I do think zero matches should be regarded a failure. slightly_smiling_face

Fair enough.

ykman/cli/oath.py
```diff
@@ -368,10 +368,13 @@ def code(ctx, show_hidden, query, single):
             if e.sw == SW.SECURITY_CONDITION_NOT_SATISFIED:
                 ctx.fail('Touch credential timed out!')
 
-    elif single:
+    elif single and len(creds) > 1:
         _error_multiple_hits(ctx, [cr for cr, c in creds])
 
-    if single:
+    elif single and len(creds) == 0:
+        ctx.fail('No credential found.')
```
emlun:
```suggestion
        ctx.fail('No matching credential found.')
```

emlun: Maybe?

PR #305, opened 2020-01-15
Improve handling of invalid PIV certificates

emlun:
Currently, both `ykman piv info` and the PIV view in ykman-qt crash if any PIV certificate slot contains data that is not a valid certificate - for example, after running `echo 'hej' | ykman piv write-object 5fc105`. These changes make the info command not crash, and allow us to also handle the same issue in the GUI.


dagheyman:
Any known scenarios except `echo 'hej' | ykman piv write-object 5fc105` where this might happen?

test/on_yubikey/cli_piv/test_misc.py
```diff
@@ -27,4 +29,19 @@ def test_write_read_object(self):
...
+                      input='Kom ihåg att du aldrig får snyta dig i mattan!')
+
+            with self.assertRaises(SystemExit):
+                ykman_cli('piv', 'export-certificate',
+                          hex(OBJ.AUTHENTICATION), '-')
+
+        def test_info_with_invalid_certificate_does_not_crash(self):
+            ykman_cli('piv', 'write-object', hex(OBJ.AUTHENTICATION), '-',
+                      '-m', DEFAULT_MANAGEMENT_KEY,
+                      input='Kom ihåg att du aldrig får snyta dig i mattan!')
```
dagheyman: Please use a more boring phrase, to avoid distraction. :)

dagheyman:
Also, python2 is still supported for the library, so would be good to test it in the CI.

test/on_yubikey/cli_piv/test_misc.py
emlun: :laughing:

emlun:
> Any known scenarios except `echo 'hej' | ykman piv write-object 5fc105` where this might happen?

Aveen and Per ran into this somehow while working on their PIV stuff. I don't know the details, but they somehow ended up with a corrupted certificate loaded into one of the slots, at which point ykman-qt was of no help to fix it.

PR #309, opened 2020-01-27
Improve OATH integration tests a bit

emlun:
- Reset YubiKey after test suite
- Test that OATH application can store 32 creds


test/on_yubikey/test_cli_oath.py
```diff
@@ -125,6 +129,16 @@ def test_oath_sha512(self):
             ykman_cli('oath', 'add', 'abba', 'abba', '--algorithm', 'SHA512')
             ykman_cli('oath', 'delete', 'abba', '-f')
 
+        def test_add_32_creds(self):
```
dagheyman: This test should not run on NEO, since it's dynamic how many credentials can be stored there.

PR #311, opened 2020-01-31
Better error handling when failing to open CCID context

fdennis:
Solves the issue when Smart Card Service on Windows is disabled and ykman qt could not open at all.

Also fixed a typo.

### DIFF
--- a/ykman/descriptor.py
+++ b/ykman/descriptor.py
@@ -160,7 +160,7 @@ def _list_drivers(transports):
                     yield dev
         except smartcard.pcsc.PCSCExceptions.EstablishContextException:
             logger.debug('Failed to establish CCID context. '
-                         'Is the pcscd service running?')
+                         'Is the pcscd/smart card service running?')
     if TRANSPORT.OTP & transports:
         for dev in open_otp():
             if dev:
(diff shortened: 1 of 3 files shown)


ykman/descriptor.py
```diff
@@ -153,9 +154,13 @@ def get_descriptors():
...
-        for dev in open_ccid():
-            if dev:
-                yield dev
+        try:
+            for dev in open_ccid():
+                if dev:
+                    yield dev
+        except smartcard.pcsc.PCSCExceptions.EstablishContextException:
+            logger.debug('Failed to establish CCID context. '
+                         'Is the pcscd service running?')
```
dagheyman: maybe we should say: `Is pcscd/smart card services running?` since it's not called pcscd on windows.

dagheyman:
Nice! Looks good, just needs some manual testing before merging.

dagheyman:
@emlun Do you know why the tests are failing? Seems unrelated. Python3 only syntax?

emlun:
Uh, that's a syntax error in the `mock` dependency, and it looks like a beta release? But yes, looks like py2-incompatible syntax. Do we care enough to restrict the dependency version or do we just drop py2 now?

dagheyman:
I think restricting the dependency version is the correct fix for now.

PR #315, opened 2020-02-14
Migrate tests to tox and pytest

emlun:
`python setup.py test` is deprecated, see
https://setuptools.readthedocs.io/en/latest/history.html#v41-5-0:

>#1878: Formally deprecated the test command, with the recommendation
that users migrate to tox.

pytest has a ["fixtures" framework][1] for parameterizing tests and preparing test environments. This allows us to get rid of most of the custom test multiplication framework, and return to conventional Python test syntax with test classes as top-level module members.

This does have the drawback that tests that don't apply to a given YubiKey are now marked as skipped instead of not appearing at all, meaning we'll always have "false positive" test skips. I believe the advantages more than make up for this, though, and we might be able to improve on the skipping mechanism in the future.

pytest also makes the test output pretter. :)

[1]: https://docs.pytest.org/en/3.0.1/fixture.html


dagheyman:
Since they to not seem to depend on each other, I'd suggest splitting this up in one PR for migrating to tox and another one for introducing pytest, so we can reason about them separately.

emlun:
Will do.

PR #316, opened 2020-02-14
Migrate to tox

emlun:
`python setup.py test` is deprecated, see
https://setuptools.readthedocs.io/en/latest/history.html#v41-5-0:

>#1878: Formally deprecated the test command, with the recommendation
that users migrate to tox.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,13 +4,15 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py27,py38
+envlist = py
 
 [testenv]
 deps =
     -r{toxinidir}/test_requirements.txt
+commands_pre =
+    python --version
 commands =
-    python -m unittest {posargs}
+    python -m unittest discover {posargs}
 passenv =
     py38: DESTRUCTIVE_TEST_YUBIKEY_SERIALS
     py38: DESTRUCTIVE_TEST_DO_NOT_PROMPT
--- a/tox.ini
+++ b/tox.ini
@@ -7,12 +7,25 @@
 envlist = py
 
 [testenv]
+# Pure software tests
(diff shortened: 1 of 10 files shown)


emlun:
Ah right, the `additional_tests()` hook [used for the integration tests](https://github.com/Yubico/yubikey-manager/blob/88642605dc45b3a1f7e50e053b01f32154777e3b/doc/integration-tests.adoc#how-it-works) is not from unittest but from setuptools, so we can't just replace `python setup.py unittest` with `python -m unittest` without also adding some code to discover and build the test suite. Migrating to pytest would fix that, so let's discuss/take care of that first. We'll probably need to close this again.

tox.ini
```diff
@@ -0,0 +1,32 @@
...
+# and then run "tox" from this directory.
+
+[tox]
+envlist = py
+
+[testenv]
+# Pure software tests
+# Invoke as `tox`, `tox -e py27` etc.
+deps =
+    -r{toxinidir}/test_requirements.txt
```
dagheyman: can the test deps be inlined in the tox file itself?

tox.ini
```diff
@@ -0,0 +1,32 @@
...
+# Tests with real YubiKeys
+# Invoke as `tox -e integration-test`
+basepython = python
+deps =
+    {[testenv]deps}
+    pytest
+commands_pre =
+    {[testenv]commands_pre}
+commands =
+    python setup.py test
```
dagheyman: Should this syntax change as well?

emlun: Yes, eventually, but right now this remains the only way to run the integration tests.

tox.ini
emlun: Yes, will do

PR #322, opened 2020-03-11
static password: add BEPO keyboard layout

dagheyman:
Port of https://github.com/Yubico/yubikey-personalization-gui/commit/e23ee929e67f13e704818cb0a3ce622231d2027e

With this change, there is no longer any keyboard layout for static passwords in yubikey-personalization-gui not also available in ykman CLI (FYI @klali ). We should direct users that wish to add more layouts to this project.

### DIFF
--- a/ykman/scancodes/__init__.py
+++ b/ykman/scancodes/__init__.py
@@ -34,10 +34,10 @@ class KEYBOARD_LAYOUT(Enum):
     MODHEX = modhex.scancodes
     US = us.scancodes
     DE = de.scancodes
-    NORMAN = norman.scancodes
     FR = fr.scancodes
     IT = it.scancodes
     BEPO = bepo.scancodes
+    NORMAN = norman.scancodes
 
 
 def encode(data, keyboard_layout=KEYBOARD_LAYOUT.MODHEX):
--- a/ykman/scancodes/__init__.py
+++ b/ykman/scancodes/__init__.py
@@ -27,7 +27,7 @@
 
 from __future__ import absolute_import
 from enum import Enum
-from . import us, de, fr, it, modhex, norman
+from . import us, de, fr, it, modhex, norman, bepo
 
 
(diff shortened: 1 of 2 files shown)


ykman/scancodes/__init__.py
```diff
@@ -27,16 +27,17 @@
 
 from __future__ import absolute_import
 from enum import Enum
-from . import us, de, fr, it, modhex, norman
+from . import us, de, fr, it, modhex, norman, bepo
```
emlun: Is it time to sort these? :smile:

ykman/scancodes/__init__.py
```diff
@@ -27,16 +27,17 @@
...
 
 class KEYBOARD_LAYOUT(Enum):
     MODHEX = modhex.scancodes
     US = us.scancodes
     DE = de.scancodes
-    NORMAN = norman.scancodes
     FR = fr.scancodes
     IT = it.scancodes
+    BEPO = bepo.scancodes
+    NORMAN = norman.scancodes
```
emlun: Is it time to sort these? :smile:

dagheyman: I did sort them from default to most obscure..

ykman/scancodes/__init__.py
dagheyman: We don't currently sort imports in anyway right? If we decide to I think we should do it across the project. I looked quickly if Black supports this, but doesn't look like it https://github.com/psf/black/issues/333

ykman/scancodes/__init__.py
emlun: Ah, I see. :)

ykman/scancodes/__init__.py
emlun: Not really, no. I've been doing it for new imports but it's indeed not uniform across the project. There are tools like `isort` for doing it automatically if we want to, but never mind for now.

PR #323, opened 2020-03-12
drivers: improve error handling and debugging

dagheyman:
- Move to lists instead of generators when dealing with the drivers, to be able to reason about the full set of drivers available. For example if we have two yubikeys reporting CCID mode enabled, but only one CCID reader visible to the OS, something might be wrong.

- Add some debug logging when the there is no drivers available for the transports asked for.

### DIFF
--- a/ykman/descriptor.py
+++ b/ykman/descriptor.py
@@ -153,27 +153,24 @@ def get_descriptors():
 
 
 def _list_drivers(transports):
+    res = []
     if TRANSPORT.CCID & transports:
         try:
-            for dev in open_ccid():
-                if dev:
-                    yield dev
+            res.extend(open_ccid())
         except smartcard.pcsc.PCSCExceptions.EstablishContextException:
             logger.debug('Failed to establish CCID context. '
                          'Is the pcscd/smart card service running?')
     if TRANSPORT.OTP & transports:
-        for dev in open_otp():
-            if dev:
-                yield dev
+        res.extend(open_otp())
     if TRANSPORT.FIDO & transports:
-        for dev in open_fido():
-            if dev:
-                yield dev
+        res.extend(open_fido())
+    return res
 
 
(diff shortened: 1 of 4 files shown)


fdennis:
I think this looks good. Perhaps the `res` list should be renamed to something like `devices`?

dagheyman:
> I think this looks good. Perhaps the `res` list should be renamed to something like `devices`?

Good point, fixed

dagheyman:
Does this looks reasonable @dainnilsson @emlun ?

dainnilsson:
I seem to recall there being a reason for having this as a generator. I don't recall the specifics, but I think when looking for a particular device we want to be able to stop once it is found, rather than open additional devices at that point. Not sure if this need was due to functional issues or just a speed optimization.

dagheyman:
> I seem to recall there being a reason for having this as a generator. I don't recall the specifics, but I think when looking for a particular device we want to be able to stop once it is found, rather than open additional devices at that point. Not sure if this need was due to functional issues or just a speed optimization.

Good call, this does indeed increase the time of for example the list command:

(first this branch then master)
```
(yubikey-manager-e7d64SWw) $ time ykman list
YubiKey 4 [OTP+FIDO+CCID] Serial: 5970500
YubiKey NEO [OTP+FIDO+CCID] Serial: 5500151

real    0m3,833s
user    0m0,319s
sys     0m0,067s
(yubikey-manager-e7d64SWw) $ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
(yubikey-manager-e7d64SWw) $ time ykman list
YubiKey 4 [OTP+FIDO+CCID] Serial: 5970500
YubiKey NEO [OTP+FIDO+CCID] Serial: 5500151

real    0m0,462s
user    0m0,316s
sys     0m0,063s
```

Maybe there is a better way to design this, will not merge this one.

PR #331, opened 2020-03-26
No kill scdaemon

dagheyman:


fdennis:
Want to make sure I don't misunderstand, but you write "Maybe a simple alias that always kills scdaemon before running ykman could work?", but doesn't that leave us with the same problem? Users would still be surprised by this I think. Do App Stores think it is fine to kill scdaemon before running ykman?

dagheyman:
> Want to make sure I don't misunderstand, but you write "Maybe a simple alias that always kills scdaemon before running ykman could work?", but doesn't that leave us with the same problem? Users would still be surprised by this I think. 

Sorry, I meant document how users can create their own alias if they are using gpg as well. Like 
```
$ alias ykman2="killall gpg-agent && ykman"
```
 or similar.

> Do App Stores think it is fine to kill scdaemon before running ykman?

No, they typically don't allow you to kill other running processes.

PR #335, opened 2020-05-03
Add --password-stdin option to ykman oath

dcepelik:
Instead of reading the password from the terminal device, read the
password from stdin. This makes it easy to pipe the password from
another command, e.g. pass.

If both -p PASSWORD and -S are provided, the password provided on stdin takes precedence.

### DIFF
--- a/ykman/cli/oath.py
+++ b/ykman/cli/oath.py
@@ -28,6 +28,7 @@
 from __future__ import absolute_import
 import click
 import logging
+import sys
 from threading import Timer
 from binascii import b2a_hex, a2b_hex
 from .util import (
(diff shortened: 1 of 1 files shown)


dagheyman:
Thanks for the contribution!

However, can't this be solved with command substitution and similar techniques? I'm hesitant to add this flag to the oath commands only, since no other ykman command (like piv, fido etc) has this option.

Example:
```
$ ykman oath --password "$(cat pw.txt)" code
```

dcepelik:
Hi @dagheyman, no problem.

> $ ykman oath --password "$(cat pw.txt)" code

Of course, that's not the issue. The issue is that providing the password on the command line is inherently insecure, as it leads to the password becoming part of the process details, i.e. visible in e.g. top or ps aux listing; and all users on the system can see it.

(There are ways to get around that, e.g. hidepid, but they don't seem to be universally supported. Furthermore, that sounds like a complicated work-around where the correct solution is not accepting the password as an argument in the first place.)

In retrospect, I've found a solution using expect (I will clean it up and paste it later), but it seems that ykman should make it easy to do the right thing.

The option should probably be added to other sub-commands where applicable, I just don't use those yet and didn't have the time to add it. I agree that's an inconsistency.

dainnilsson:
Tha handling around this has been improved in ykman 4.0. As before, if no password is provided one will be prompted for on stdin. Now, piping the password to the command should function as expected, making this flag unneeded.

PR #348, opened 2020-06-12
add Bandit security scanner and suppress some of the warnings

dagheyman:
This adds [Bandit](https://github.com/PyCQA/bandit) integration as a pre-commit hook, which also runs on GitHub actions.

It also either fixes or suppresses some of the warnings:
- Add logging to empty Exception blocks
- Suppress warnings for use of SHA1 in a HMAC contexts
- Suppress warning for use of ECB in relation to the PIV management key
- Drop one use of assert, convert one use of assert to a regular check
- Suppress warning for use of subprocess module, HTTPSConnection, the word "secret" in some constants

To run bandit manually:

`$ bandit -r .`

Note that exclude directories are defined in the `.bandit` config file as well as in the pre-commit config.

I opted for suppressing warnings directly in the code with the `# nosec` syntax, since we already do that for flake8 exceptions and it provides an opportunity to explain why something is suppressed as well.

### DIFF
--- a/ykman/driver_ccid.py
+++ b/ykman/driver_ccid.py
@@ -300,14 +300,10 @@ def kill_scdaemon():
                 killed = True
     except ImportError:
         # Works for Linux and OS X.
-        pids = subprocess.check_output(  # nosec
-            "ps ax | grep scdaemon | grep -v grep | awk '{ print $1 }'",
-            shell=True).strip()
-        if pids:
-            for pid in pids.split():
-                subprocess.call(['kill', '-9', pid])  # nosec
+        return_code = subprocess.call(  # nosec
+            ['/usr/bin/pkill', '-9', 'scdaemon'])
+        if return_code == 0:
             killed = True
-
     if killed:
         time.sleep(0.1)
     return killed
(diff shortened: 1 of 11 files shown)


ykman/driver_ccid.py
```diff
@@ -300,12 +300,12 @@ def kill_scdaemon():
...
     except ImportError:
         # Works for Linux and OS X.
-        pids = subprocess.check_output(
+        pids = subprocess.check_output(  # nosec
             "ps ax | grep scdaemon | grep -v grep | awk '{ print $1 }'",
             shell=True).strip()
         if pids:
             for pid in pids.split():
-                subprocess.call(['kill', '-9', pid])
+                subprocess.call(['kill', '-9', pid])  # nosec
```
nevun, line 308: I think it would be better to replace both of these subprocess calls with one:

subprocess.call(["/usr/bin/pkill", "-9", "scdaemon"])                                                                                                   

This works on macOS and on Linux, /usr/bin/pkill is the correct path on macOS, OpenBSD, FreeBSD, ArchLinux, Debian, Ubuntu and Fedora (yes, I checked all of them now).  

This reduced the number of warnings from bandit from 5 to 2.

dagheyman: Agreed, fixed.

PR #353, opened 2020-08-07
Improve listing of invalid/unsupported certificates in piv

dainnilsson:
Avoid breaking the loop in `ykman piv info` if one certificate can't be shown.


dainnilsson:
Already fixed in main.

Issue #361, opened 2020-10-03
AttributeError: '_Ed25519PrivateKey' object has no attribute 'key_size'


yawnt:
- YubiKey Manager (ykman) version: 3.1.1
- How was it installed?: ArchLinux AUR
- Operating system and version: ArchLinux (5.8.10)
- YubiKey model and version: Yubikey 5 NFC (5.2.6)
- Bug description summary: ykman throws AttributeError when importing .pem key

Steps to reproduce
---

```
$ openssl genpkey -algorithm Ed25519 -out ca.pem
$ ykman piv import-key 9a ca.pem
```

Expected result
---

Key is imported into slot 9A

Actual results and logs
---

```
$ ykman piv import-key 9a ca.pem
Enter PIN: 
Traceback (most recent call last):
  File "/usr/bin/ykman", line 33, in <module>
    sys.exit(load_entry_point('yubikey-manager==3.1.1', 'console_scripts', 'ykman')())
  File "/usr/lib/python3.8/site-packages/ykman/cli/__main__.py", line 273, in main
    cli(obj={})
  File "/usr/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3.8/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/lib/python3.8/site-packages/ykman/cli/piv.py", line 411, in import_key
    _check_key_size(ctx, controller, private_key)
  File "/usr/lib/python3.8/site-packages/ykman/cli/piv.py", line 982, in _check_key_size
    if (private_key.key_size == 1024
AttributeError: '_Ed25519PrivateKey' object has no attribute 'key_size'
```

[...]

PR #371, opened 2021-01-04
Improve type-checking for imported keys.

EliZucker:
This PR adds reasonable output for cases like those seen in [issue #361](https://github.com/Yubico/yubikey-manager/issues/361).

I modified `_check_key_size()` to only verify key size if applicable, and then added additional checking for the call to `controller.import_key()` — which will throw an exception when a certain key/algorithm is not supported. A similar try/except already exists in `generate_key()`'s call to `controller.generate_key()`, but I couldn't find a reason it shouldn't exist here as well.

### DIFF
--- a/ykman/cli/piv.py
+++ b/ykman/cli/piv.py
@@ -410,11 +410,14 @@ def import_key(
     _check_touch_policy(ctx, controller, touch_policy)
     _check_key_size(ctx, controller, private_key)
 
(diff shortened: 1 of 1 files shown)


dainnilsson:
Thank you for the contribution! The code around this has changed a lot in the `next` branch, and the error message there has been improved.

PR #385, opened 2021-02-11
cryptography.utils.int_from_bytes deprecated, use int.from_bytes

jay0lee:
The cryptography library deprecated cryptography.utils.int_from_bytes in https://github.com/pyca/cryptography/pull/5609. Use the builtin int.from_bytes instead.


dainnilsson:
Already implemented on `next`.

PR #392, opened 2021-02-21
add poetry-based dev instrucions

maxthomas:
hey there, big fan of the company, tools, and software :)

looks like there's no longer a `setup.py` so it seems the dev documentation about `pip install -e .` is bit rotted. i added some basic instructions for the new `poetry` based setup for those looking to hack on the project, like me - if there's a better way that y'all have moved on to, that'd be great as well!

thanks for your time and efforts on this project!

### DIFF
--- a/doc/development.adoc
+++ b/doc/development.adoc
@@ -3,6 +3,8 @@
 === Install dependencies
 
 It's assumed a Python 3 environment with pip is installed.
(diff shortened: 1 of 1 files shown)


dainnilsson:
Thanks! I could have sworn I had updated that file...

PR #411, opened 2021-04-22
Patch missing FreeBSD ykman/hid Backend support

egypcio:
by merging this request, we:

    - set 'backend' to be 'None' if we are on FreeBSD;
      - there's no official support for this platform;
    - avoid raising an exception on this special case;
    - replace `backend.list_devices` by `lambda: []` only for this platform;
    - make it possible to build, install and use YubiKey Manager on FreeBSD.

this patch was tested and suggested by FreeBSD developers (all credits to nwhitehorn@);
please refer to https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=254580 for more information.

### DIFF
--- a/ykman/hid/__init__.py
+++ b/ykman/hid/__init__.py
@@ -47,10 +47,12 @@ elif sys.platform.startswith("freebsd"):
 else:
     raise Exception("Unsupported platform")
 
+
 if backend is not None:
     list_otp_devices: Callable[[], List[OtpYubiKeyDevice]] = backend.list_devices
 else:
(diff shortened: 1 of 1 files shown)


dainnilsson:
Thank you. I see what the goal is here, but I'd rather not have this type of "special handling" for FreeBSD that causes it to behave in a different way than other "unsupported" platforms. I've opened #414 to solve the underlying problem instead, hopefully that will take care of your use-case.

dainnilsson:
This has now been addressed in 711b13a3564b13aea57c71edf4b96553f463e2b5 hopefully making this PR obsolete. Please comment on #414 if this doesn't fully resolve your issue.

PR #423, opened 2021-06-21
Generate key if generate is chosen with TOTP

jamiehankins:
Currently, for the "otp chalresp" command, the tool allows the user to specify
both generate and TOTP, but it ignores the generate and prompts the user for a
key, anyhow.

This change checks generate by itself and generates the key if it's true. If
TOTP is chosen, it outputs the generated key as Base32, since that's what it
prompts the user for if generate isn't chosen.

If generate and TOTP are not meant to be used together, then instead of this
change, we should print an error message if both are chosen.


dainnilsson:
I think the original intent was to not have `generate` and `totp` be used together, but I think the approach of allowing it and outputting base32 makes sense, so I say let's keep it! Thanks!

PR #425, opened 2021-07-04
Add REPL command

meeuw:
I'd like to use ykman without restarting it everytime, it is possible to import `import ykman.cli.__main__` and call `ykman.cli.__main__.cli.main` but then ykman messes with your log handlers etc. I figured it might be better to add a REPL option to ykman so ykman keeps running and you can send queries over stdin.

This script can be used as a proof of concept for this PR:
```python
import subprocess
import time
import json
import sys

p = subprocess.Popen(
    ["ykman", "repl"],
    stdin=subprocess.PIPE,
    stdout=subprocess.PIPE,
    universal_newlines=True,
    bufsize=1,
)

while 1:
    p.stdin.write(json.dumps(["oath", "accounts", "code"]) + "\n")
    print(json.loads(p.stdout.readline())["stdout"][0])
    time.sleep(1)
```

This is a bit of a hack and might require a recent Python version (I'm using 3.9). For my use case JSON input/output is very useful, as a stand alone REPL python-prompt-toolkit might be better (or another readline implementation).

Please let me know what's required to get this merged in main.

### DIFF
--- a/ykman/cli/__main__.py
+++ b/ykman/cli/__main__.py
@@ -60,6 +60,9 @@ import ctypes
 import time
 import sys
 import logging
+import json
+import io
+import contextlib
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


emlun:
Sorry, I don't think we want to commit to supporting something like this at this time. But you can still do this as a standalone application:

```python
#!/usr/bin/python3

import contextlib
import io
import json
from ykman.cli.__main__ import cli

"""
Enter commands as REPL (formatted as JSON)
"""


while 1:
    line = input()
    if not line:
        break
    stdout = io.StringIO()
    stderr = io.StringIO()
    try:
        with contextlib.redirect_stderr(stderr), contextlib.redirect_stdout(stdout):
            cli.main(args=json.loads(line))
    except SystemExit:
        pass
    print(
        json.dumps(
            {
                "stdout": stdout.getvalue().splitlines(),
                "stderr": stderr.getvalue().splitlines(),
            }
        ),
        flush=True,
    )
```

Just note that we don't generally promise to keep text output stable for programmatic consumption, with a few exceptions like the `--single` option to `oath accounts code`.

meeuw:
Thanks a lot for your response @emlun! I figured that already. Very famliar snippet :-)

https://github.com/meeuw/yubikey-manager-lib/blob/master/yubikey_manager_repl.py

PR #443, opened 2021-07-23
Add shell completion for --device and --reader

### LINKED ISSUE #439: [Feature Request] Add autocomplete for `--device` and `--reader` parameters.
- YubiKey Manager (ykman) version: 4.0.3
- How was it installed?: from PyPI.org using pip
- Operating system and version: Linux Ubuntu 21.04
- YubiKey model and version: Yubikey 5C Nano (5.4.3)
- Bug description summary: Not being able to autocomplete device serial number (required when multiple yubikeys are plugged in)

Steps to reproduce
---
1. Make sure you have auto-completion for the `ykman` command installed in your shell (see [README](https://github.com/Yubico/yubikey-manager#shell-completion) for instructions)
2. Make sure at least one Yubikey is plugged in (you can check it with `ykman list`
3. Type in your console `ykman --device=` or `ykman --reader=` and press the tab key
4. Observe the results

Expected result
---

The auto-completion of the Yubikey serial ID (when `--device=` is used) or supported readers (when `--reader=` is used) should appear

Actual results and logs
---

Nothing happens

Other info
---

None

emlun:
(Experimental/speculative)

Fixes #439.

In bash, the suggestions will simply be the available serial numbers. In more capable shells like zsh or fish, the suggestions will be annotated with the same device descriptions as shown by `ykman list`.

Example:
2021-07-23-20:47:57_1390x472

Marking this as a draft since this would bump the Click dependency to minimum version 8, and we might not want to require that. Version 7 also supports a variant of this, but without the "suggestion help" feature of the more capable shells.


dainnilsson:
Looks interesting. I think we could fairly easily degrade on older versions of Click rather than require Click 8.

The thing that concerns me about this is that enumerating attached YubiKeys with serial numbers is a pretty slow/heavy operation, which also interrupts anything those YubiKeys are currently doing (eg. killing scdaemon, etc.). I'm a bit worried that this would be unexpected to users.

emlun:
Yeah, good point. I experimented a bit more with an explicit opt-in step... try it out. :slightly_smiling_face: I haven't tried it in zsh or fish, but it works pretty well in bash at least. But in the end this might be too much complexity for too little gain.

kurtmckee:
Came to this repo in hopes that bash completion had been implemented, including commands, subcommands, and device numbers. Probably the Click 8 dependency isn't an important consideration 2 years on.

emlun:
Right you are, the [dependency in `pyproject.toml`](https://github.com/Yubico/yubikey-manager/blob/51a7ae438c923189788a1e31d3de18d452131942/pyproject.toml#L34) on `main` now already specifies 8 as the minimum Click version. So perhaps it's time to revisit this @dainnilsson?

dainnilsson:
The main issue with this wasn't the Click 8 requirement, it was the operation requiring interrupting several attached devices and being slow. Unless that changes this needs to be locked away behind a secret flag, and I don't really see it as very useful in that case. I could probably be convinced otherwise though.

emlun:
Yes, that is still the case, and the feature as implemented is disabled unless the environment variable `_YKMAN_EXPERIMENTAL_COMPLETE_DEVICE` exists. Sounds like at least @kurtmckee would still use it? I would too, for what it's worth.

dainnilsson:
After some internal discussion I've pushed a new commit with an alternative approach:

Any time devices are enumerated as part of standard usage (most ykman commands will do this) the list of devices is cached in an LRU cache, for use with tab-completion later on. This means that the devices that will complete aren't necessarily present at the moment, but have been "recently" present. The benefit of this is that we never interrupt devices specifically for tab completing purposes, and that it is much faster to actually invoke tab completion. One exception is when running a command without specifying a serial when multiple devices are attached: Here I added an explicit update to the cache as it is very likely the user will immediately try to re-run the command with a serial specified.

The limit on the cache is currently set to 3 devices, which is probably too low, to make testing the behavior a bit easier.

Issue #476, opened 2022-01-02
Manage openpgp data


kilimnik:
It is currently impossible to manage and retrieve data via the ykman cli/library.
By data I am referring to the name, login data, language preference, sex and url

It would be useful, if that functionality would be added. That would make managing the data via python easier.

NukeManDan:
dup of #495 I think

PR #477, opened 2022-01-02
Manage openpgp data

kilimnik:
I implemented my feature request #476.

For some reason the retieving of the name, language preference and sex always failed when trying to request value directly. That's the reason I implemented it via the constructed DO.


kilimnik:
I do not think the data is useless. First of all, it helps to differentiate between Yubikeys, if you have multiples.
Second and the most notable, is that the url field is needed to use the gpg-agent. It uses the url field to pull the public gpg key from a keyserver.

Issue #478, opened 2022-01-03
Manage openpgp force signature pin


kilimnik:
It is currently impossible to manage the openpgp force signature pin setting via ykman cli/library.

It would be useful, if that functionality would be added. That would make managing the force signature pin via python easier.

NukeManDan:
Duplicate #495 I think

dainnilsson:
ykman 5 has now been released which includes this functionality.

PR #479, opened 2022-01-03
add force signature pin capabilities

kilimnik:
I implemented my feature request #478.

### DIFF
--- a/ykman/cli/openpgp.py
+++ b/ykman/cli/openpgp.py
@@ -28,7 +28,13 @@
 import logging
 import click
 from ..util import parse_certificates, parse_private_key
-from ..openpgp import OpenPgpController, KEY_SLOT, TOUCH_MODE, get_openpgp_info
+from ..openpgp import (
+    OpenPgpController,
+    KEY_SLOT,
+    TOUCH_MODE,
+    PIN_POLICY,
+    get_openpgp_info,
+)
 from .util import (
     CliFail,
     click_force_option,
@@ -235,6 +241,33 @@ def change_admin(ctx, admin_pin, new_admin_pin):
         )
 
     controller.change_admin(admin_pin, new_admin_pin)
+
+
+@access.command("set-signature-policy")
+@click.argument("policy", metavar="POLICY", type=EnumChoice(PIN_POLICY))
+@click.option("-a", "--admin-pin", help="Admin PIN for OpenPGP.")
+@click.pass_context
+def set_signature_policy(ctx, policy, admin_pin):
+    """
+    Set Signature PIN policy.
+
+    \b
+    POLICY  Signature PIN policy to set (always, once).
+
+    The Signature PIN policy is used to control whether the PIN is
+    always required when using the Signature key, or if it is required
+    only once per session.
+    """
+    controller = ctx.obj["controller"]
+
+    if admin_pin is None:
+        admin_pin = click_prompt("Enter Admin PIN", hide_input=True)
+
+    try:
+        controller.verify_admin(admin_pin)
+        controller.set_signature_pin_policy(policy)
+    except Exception:
+        raise CliFail("Failed to set new Signature PIN policy")
 
 
 @openpgp.group("keys")
--- a/ykman/openpgp.py
+++ b/ykman/openpgp.py
@@ -108,6 +108,18 @@ class TOUCH_MODE(IntEnum):  # noqa: N801
 
 
 @unique
+class PIN_POLICY(IntEnum):  # noqa: N801
+    ONCE = 0x00
+    ALWAYS = 0x01
+
+    def __str__(self):
+        if self == PIN_POLICY.ONCE:
+            return "Once"
+        elif self == PIN_POLICY.ALWAYS:
+            return "Always"
+
+
+@unique
 class INS(IntEnum):  # noqa: N801
     GET_DATA = 0xCA
     GET_VERSION = 0xF1
@@ -375,6 +387,17 @@ class OpenPgpController(object):
         data = self._get_data(DO.PW_STATUS)
         return PinRetries(*data[4:7])
 
+    def get_signature_pin_policy(self):
+        data = self._get_data(DO.PW_STATUS)
+
+        return PIN_POLICY(data[0])
+
+    def set_signature_pin_policy(self, pin_policy):
+        """Requires Admin PIN verification."""
+        data = struct.pack(">B", pin_policy)
+
+        self._put_data(DO.PW_STATUS, data)
+
     def _block_pins(self):
         retries = self.get_remaining_pin_tries()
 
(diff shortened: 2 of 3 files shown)


kilimnik:
@dainnilsson it would be great, if you could review this and also my other pull requests, as they have been open for almost two months now.

dainnilsson:
Sorry for taking some time on this, the PRs are quite substantial and I want to be able to dedicate enough time to go over them in detail. Unfortunately I've been busy with other matters, but I do hope to be able to look into this soon (next couple of weeks). Again, sorry for the delay, and thank you for your patience.

dainnilsson:
Thank you for your contributions!

Since you have 3 PRs open, let me address them all here first:

I think the PIN additions (this PR as well as the one adding support for changing the PINs) make sense in the context of this project. I'm happy to work with these PRs and get them merged. The third one, for managing user data, is one that I'm not convinced about. We've seen little or no practical use of these fields in general, and even less need for the ability to edit them from ykman. For now, this isn't something we wish to support in this project.

I think it would make sense to merge #473 first, followed by this PR, so I'll start there.

dainnilsson:
I've been thinking a bit around what to call things here. On the one hand I want it to be clear that this is the "same thing" that `gnupg` calls `Signature PIN` with values `forced` and `not forced`. On the other hand I want the commands and arguments to be consistent with the rest of `ykman`. Additionally, `not forced` also has the problem of having a space in it, so it doesn't work well as a command line parameter.

Here's what I'm thinking:
`set-pin-policy POLICY`, where `POLICY` is `always` or `once`.

The `set-` prefix is one we often use in these cases, and we use the "PIN policy" terminology for PIV, where we also have the `always` and `once` values. I think `always` and `once` are slightly more clear than `forced` and `not forced`, but mainly don't have the issue with space, and are already used elsewhere in the application.

Of course, `set-signature-pin-policy` would be more clear in *which* PIN policy this changes, but since that would make the command so long, and GPG really only has one PIN policy, I think it's good enough as-is. The `--help` output should make it clear what it does anyway, in case of confusion.

I would then also move reading of the value instead to the `ykman openpgp info` output. Something along the lines of `Signature PIN policy: Once|Always`.

For these changes, I would introduce a PIN_POLICY IntEnum (with values `ONCE==0x00`, `ALWAYS=0x01`) similar to the existing `TOUCH_MODE` one (in ykman/openpgp.py), which would be used as argument to a `controller.set_signature_pin_policy(policy)` and as a return type of `controller.get_signature_pin_policy()`.

What do you think?

kilimnik:
I do like the change to `POLICY` beeing `always` and `once`. I also think `set-signature-pin-policy` would make it more clear. What do you think about abbreveating it to `set-sig-pin-policy`. There are already examples of rather long commands like in PIV there is `change-management-key`.

dainnilsson:
We try to avoid abbreviating words in function and command names. How about `set-signature-policy`? Since this is under the `access` group, it should already be somewhat clear that this relates to PIN (and the --help text would clarify this further). I suppose even `set-signature-pin-policy` could work even though it is quite long, since it's not a command you would typically run very often.

kilimnik:
I think `set-signature-policy` is fine. I will start working on the changes as soon as possible

ykman/cli/openpgp.py
```diff
@@ -107,6 +108,16 @@ def info(ctx):
...
 
+    list: List[Tuple[str, str]] = []
+    list.append(("Signature pin policy", controller.get_signature_pin_policy()))
+
+    column_left_lengths = [len(x[0]) for x in list]
+    max_column_left = max(column_left_lengths)
+
+    output = map(lambda x: f"{x[0].ljust(max_column_left)}\t{x[1]}", list)
+
+    click.echo("\n".join(output))
```
dainnilsson, line 119: I think it would make sense for this to be added to the output of `get_openpgp_info` (in ykman/openpgp.py) and just get printed here as the result of that. There you also won't have to do any formatting since `pretty_print` will take care of that.

The word "PIN" should be capitalized.

ykman/cli/openpgp.py
```diff
@@ -237,6 +248,34 @@ def change_admin(ctx, admin_pin, new_admin_pin):
...
+@click.option("-a", "--admin-pin", help="Admin PIN for OpenPGP.")
+@click.pass_context
+def set_signature_policy(ctx, policy, admin_pin):
+    """
+    Set signature pin policy.
+
+    \b
+    POLICY  Signature pin policy to set (always, once).
+
+    The signature pin policy is used to control whether the pin is
```
dainnilsson, line 262: Capitalize "PIN".

ykman/cli/openpgp.py
```diff
@@ -237,6 +248,34 @@ def change_admin(ctx, admin_pin, new_admin_pin):
...
+
+    if admin_pin is None:
+        admin_pin = click_prompt("Enter Admin PIN", hide_input=True)
+
+    try:
+        controller.verify_admin(admin_pin)
+        controller.set_signature_pin_policy(policy)
+    except Exception as e:
+        logger.debug("Failed to set new signature pin policy", exc_info=e)
+        raise CliFail("Failed to set new signature pin policy")
```
dainnilsson, line 276: Capitalize "PIN". The `logger.debug` line is not needed, the `CliFail` stacktrace will include the exception that caused it automatically.

ykman/openpgp.py
```diff
@@ -107,6 +107,18 @@ def __str__(self):
             return "Cached (fixed)"
 
 
+@unique
+class POLICY(IntEnum):  # noqa: N801
```
dainnilsson, line 111: I think this should be renamed to `PIN_POLICY` to clarify. The argument to `set-pin-policy` should still be named `policy`.

ykman/openpgp.py
```diff
@@ -375,6 +387,20 @@ def get_remaining_pin_tries(self):
         data = self._get_data(DO.PW_STATUS)
         return PinRetries(*data[4:7])
 
+    def get_signature_pin_policy(self):
+        data = self._get_data(DO.PW_STATUS)
+
+        if bool(data[0]):
+            return POLICY.ONCE
+        else:
+            return POLICY.ALWAYS
```
dainnilsson, line 396: From what I recall, the standard defines this byte as explicitly having the valid values `0x00` and `0x01`, in which case you can just `return POLICY(data[0])`.

ykman/openpgp.py
```diff
@@ -375,6 +387,20 @@ def get_remaining_pin_tries(self):
...
+
+    def set_signature_pin_policy(self, status):
+        """Requires Admin PIN verification."""
+        data = struct.pack(">B", status == POLICY.ONCE)
```
dainnilsson, line 400: `status` should be renamed to `pin_policy`. The equals-check is not needed, you can just use `struct.pack(">B", pin_policy)`.

dainnilsson (changes requested):
Mainly stylistic changes requested. Functionally I think things look good. Thanks!

dainnilsson (approved):

dainnilsson:
Thanks! This has now been merged into #494.

PR #484, opened 2022-02-02
Use secrets module instead of urandom.

createyourpersonalaccount:
The old code used `os.urandom()` to generate random bytes. The new
source code uses `secrets.token_bytes()` instead. The secrets module is
preferred for cryptographic applications.

### DIFF
--- a/tests/device/cli/piv/test_read_write_object.py
+++ b/tests/device/cli/piv/test_read_write_object.py
@@ -1,4 +1,4 @@
-import os
+import secrets
 
 from cryptography.hazmat.primitives import serialization
 from ....util import generate_self_signed_certificate
@@ -44,7 +44,7 @@ class TestReadWriteObject:
         assert data == output_data
 
     def test_read_write_read_is_noop(self, ykman_cli):
-        data = os.urandom(32)
+        data = secrets.token_bytes(32)
 
         ykman_cli(
             "piv",
@@ -79,7 +79,7 @@ class TestReadWriteObject:
         assert output2 == data
 
     def test_read_write_aliases(self, ykman_cli):
-        data = os.urandom(32)
+        data = secrets.token_bytes(32)
 
         with io.StringIO() as buf:
             with contextlib.redirect_stderr(buf):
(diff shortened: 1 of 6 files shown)


dainnilsson:
Why is the `secrets` module preferred? What improvement does this give?

createyourpersonalaccount:
@dainnilsson 

The `urandom` call has gone through various revisions in Python3, increasing security guarantees. The documentation says,

> The returned data should be unpredictable enough for cryptographic applications, though its **exact quality depends on the OS implementation**.

The secrets module has the following description,

> The secrets module is used for generating cryptographically strong random numbers suitable for managing data such as passwords, account authentication, security tokens, and related secrets.

It is available since Python 3.6 and Python 3.6 or above is already required by `pyproject.toml`. Using this module adds clarity of intention (cryptographic application) and future-proofs the implementation in case `urandom()` falls back on strength. It may also increase overall security for certain versions and implementations, but should not decrease it.

createyourpersonalaccount:
I've also checked that CPython and PyPy both just call `urandom()`. For CPython I checked all versions with `git log -L :token_bytes:Lib/secrets.py v3.5.1..v3.9.0a6` while for PyPy I downloaded a snapshot of pypy3.6. Thus this change is also backwards compatible.

The history of the `secrets` module comes from a discussion between Guido van Rossum and Theo de Raadt where they talked about a common newcomer mistake of using a non-crypto rng for crypto applications, and wanted to have *most secure* by default instead. The `os.urandom()` option was available back then, but people instead used `random.random()`. The idea of having a userland cprng implementation using OpenBSD's `arc4random()` was thrown around but ultimately people converged to using the default OS cprng. The `secrets` module is the effort to have an obvious secure solution available to newcomers.

When I made this commit, I didn't really believe that I was correcting a mistake; this is just for style points...

dainnilsson:
Thanks. I agree that the `secrets` module may be more clear in its purpose than `os.urandom`, but I don't see any other benefit to it, at least not now. From what I can tell, `os.urandom` is much more widely used. Even [cryptography.io](https://cryptography.io/en/latest/random-numbers/) (which we rely on heavily in this project) recommends using `os.urandom`, with a very short note at the bottom mentioning `secrets` as an alternative. I'll reconsider in the future if the switch can offer us some more clear benefit, but for now I'll close this.

Issue #482, opened 2022-01-31
Multiple YubiKeys with same HMAC-SHA1 challenge response mode


createyourpersonalaccount:
- YubiKey Manager (ykman) version: 4.0.8-dev0
- How was it installed?: dnf package manager
- Operating system and version: Fedora linux 5.15.16-200.fc35.x86_64
- YubiKey model and version: YubiKey 5 NFC firmware version 5.4.3
- Bug description summary: No option to program two yubikeys with same key for HMAC-SHA1 challenge response mode.
 
Steps to reproduce
---

I have two YubiKeys. I use `ykman --device XXX otp chalresp -g 2` then attempt to copy the key with `ykman --device YYY otp chalresp 2`. The second command requires a key; if I copy/paste the previous secret key it won't work. I've also tried to convert the hex key into binary data and feed `ykman` the binary data via a pipe but I get a utf8 encoding error. 

Expected result
---

I was hoping that the pipe `ykman --device YYY otp chalresp 2 < mykey` would work; or that there would be a way to feed a hex key.

Actual results and logs
---

```
$ cat mykey | ykman otp chalresp -t 2
Error: 'utf-8' codec can't decode byte 0xc7 in position 0: invalid continuation byte
```

Other info
---

If this is a missing feature for `ykman`, is there interest in developing it? I could give it a go.

dainnilsson:
> The second command requires a key; if I copy/paste the previous secret key it won't work.

This should work. What happens when you try it?

createyourpersonalaccount:
@dainnilsson You are right, I have just tested that this works. The way I tested it was by feeding the same message into the challenge-response calculation:

```
$ ykman --device XXX otp calculate 2 00 # send the 00 message
<output>
$ ykman --device YYY otp calculate 2 00
<same output>
```

My bad! Thank you for the help. I came here from an [XY problem](https://en.wikipedia.org/wiki/XY_problem) with KeePassXC, assuming that my issue was my inability to program the same key into the hmac-sha1 mode...

PR #485, opened 2022-02-02
Improve description of cli chalresp KEY argument.

createyourpersonalaccount:
When I used `ykman` I was confused about programming two yubikeys with
the same key in hmac-sha1: I didn't know if the key was read as a string
or in hex. I'm clarifying the description to explain that the key may be
either hex or base32. See also issue #482.

### DIFF
--- a/ykman/otp.py
+++ b/ykman/otp.py
@@ -34,7 +34,6 @@ from enum import Enum
 from http.client import HTTPSConnection
 from typing import Iterable
 
-import re
 import json
 import struct
 import random
@@ -166,11 +165,9 @@ def generate_static_pw(
 
 def parse_oath_key(val: str) -> bytes:
     """Parse a secret key encoded as either Hex or Base32."""
-    val = val.upper()
-    if re.match(r"^([0-9A-F]{2})+$", val):  # hex
+    try:
         return bytes.fromhex(val)
-    else:
-        # Key should be b32 encoded
+    except ValueError:
         return parse_b32_key(val)
 
 
--- a/ykman/cli/otp.py
+++ b/ykman/cli/otp.py
@@ -532,7 +532,7 @@ def static(ctx, slot, password, generate, length, keyboard_layout, no_enter, for
     "--totp",
     is_flag=True,
     required=False,
-    help="Use a base32 encoded key for TOTP credentials.",
+    help="Use a base32 encoded key (optionally padded) for TOTP credentials.",
 )
 @click.option(
     "-g",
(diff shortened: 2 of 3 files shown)


dainnilsson:
Thanks!

The `--totp` flag is the method of parsing key as base32, and we will likely begin enforing that soon (as in, always parsing the key as hex unless --totp is specified). I'd prefer id the description of KEY better reflected this. Something like "A key given in hex (or base32, if --totp is specified)" to prevent new users from relying on this ambiguity. Maybe the description for `--totp` is a better place to point out that padding is optional?

This is something that wasn't introduced by this PR, but I noticed there's an inconsistency with capitalization of Base32 on line 571 (we use lowercase b everywhere else). Would you mind fixing that while you're at it?

createyourpersonalaccount:
@dainnilsson By implementing your suggested changes, `parse_oath_key()` behaves differently from documentation. I decided to update `parse_oath_key()` accordingly. See the new force-pushed commit.

dainnilsson:
The change to only allow hex when `--totp` is not specified is the change I mentioned earlier, that we want to make "soon". I don't want that change into `main` right now, as it would break things for anyone relying on this behavior (hopefully no one). I would be fine with either of the following options:

1. Revert the change to `parse_oath_key`, keeping the existing behavior. I do not think that this makes it behave in a way that isn't consistent with the documentation. The help states that the key will be parsed as hex, and that is what the function does, for all valid hex inputs. The fact that it falls back to parsing base32 is undocumented, but doesn't conflict with the documentation in my opinion.

2. Keep the change of failing on invalid hex, but retarget this PR for the `next` branch instead of `main`. This may mean it takes a bit longer before this PR goes into a release. In this case, I would prefer it if we just remove `parse_oath_key` altogether, and just use `bytes.fromhex` directly. Note that uncaught ValueErrors will cause the CLI to fail automatically, so it only needs to be caught (in ykman/cli/otp.py) if the error message needs to be changed.

createyourpersonalaccount:
@dainnilsson Okay, I'm going with 1. then, see newest commit.

ykman/otp.py
```diff
@@ -165,12 +164,10 @@ def generate_static_pw(
 
 
 def parse_oath_key(val: str) -> bytes:
-    """Parse a secret key encoded as either Hex or Base32."""
-    val = val.upper()
-    if re.match(r"^([0-9A-F]{2})+$", val):  # hex
+    """Parse a secret key encoded encoded as either Hex or Base32."""
```
dainnilsson, line 167: Duplicated word: "encoded encoded".

dainnilsson:
Looks great, thanks! One minor comment fix, then I think it's good to go.

ykman/otp.py
createyourpersonalaccount: Fixed.

createyourpersonalaccount:
Should be ready.

dainnilsson:
Thanks! Merged.

PR #511, opened 2022-06-19
Handle APPLET_SELECT_FAILED in smartcard applet select

StarGate01:
The `APPLET_SELECT_FAILED` ISO7816 APDU error code (https://docs.oracle.com/javacard/3.0.5/api/javacard/framework/ISO7816.html#SW_APPLET_SELECT_FAILED) is returned by some card environments instead of `FILE_NOT_FOUND` . For example, https://github.com/licel/jcardsim (https://jcardsim.org/) behaves this way when an applet fails to load or was not found.

Without this change, ykman is unable to properly enumerate the installed applets in this environment, which leads to an early abort of any command.

### DIFF
--- a/yubikit/core/smartcard.py
+++ b/yubikit/core/smartcard.py
@@ -78,6 +78,7 @@ class SW(IntEnum):
     FILE_NOT_FOUND = 0x6A82
     NO_SPACE = 0x6A84
     REFERENCE_DATA_NOT_FOUND = 0x6A88
+    APPLET_SELECT_FAILED = 0x6999
     WRONG_PARAMETERS_P1P2 = 0x6B00
     INVALID_INSTRUCTION = 0x6D00
     COMMAND_ABORTED = 0x6F00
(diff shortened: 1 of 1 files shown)


dainnilsson:
Could you explain the use case for using ykman with jcardsim (or anything else that isn't a YubiKey, really)?

StarGate01:
I am working on understanding the underlying protocols and creating integrations with other software, and emulating Javacard applets for fine-grained debugging is a very useful tool. In addition, dealing with real hardware (Yubikeys) is often inconvenient for automated testing and CI systems, where everything runs on remote VMs. I use ykman to initialize the emulated applets, and then perform regression / functional testing of my other software.

StarGate01:
Thank you for understanding! This small patch indeed significantly improves my workflow. The `APPLET_SELECT_FAILED` code is standard-compliant, and as you wrote the patch has minimal impact.

dainnilsson:
This has now been merged both into `main` as well as `next`.

PR #533, opened 2022-12-19
Remove trailing whitespace from info table

bittner:
The `ykman info` command generates tabular information about an attached device. The produced table has unnecessary trailing whitespace on each line. This change removes the trailing whitespace.


dainnilsson:
Thanks! This has now been merged into `main`.

PR #567, opened 2023-08-16
Add API documentation

elibon99:
This adds documentation to `yubikit` and `ykman`. Furthermore, Sphinx is added as a dependency to generate the API documentation.


yubikit/openpgp.py
```diff
@@ -955,6 +988,8 @@ def _pad_message(attributes, message, hash_algorithm):
 
 
 class OpenPgpSession:
+    """A session with the OpenPGP appliation."""
```
dainnilsson, line 991: Typo in application

elibon99: Pushed a commit that resolves this and a couple of other typos I found.

dainnilsson (approved):

PR #575, opened 2023-11-13
Add a Dependabot config to update GitHub action versions

kurtmckee:
This PR introduces the following change:

* Add a Dependabot configuration to update GitHub action versions via PRs.

There are several actions that are no longer current (`actions/checkout@v3` is now at v4 for example, and `actions/upload-artifact@v1` is now at v3). Rather than submit a PR to update these just once, the new Dependabot config will make sure they stay up-to-date on a regular basis.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"


emlun:
See for example: https://github.com/Yubico/java-webauthn-server/issues?q=label%3Agithub_actions+author%3Aapp%2Fdependabot

(Dependabot config [here](https://github.com/Yubico/java-webauthn-server/blob/3c74bb57f4abe94488a854fdaaf3acd5b469603f/.github/dependabot.yml#L15-L18))

kurtmckee:
Right, exactly. The PRs in this configuration will be similar, but Dependabot will batch the updates together. [Here is a representative example](https://github.com/kurtmckee/feedparser/pull/406) of that batching.

dainnilsson:
Thanks!

PR #598, opened 2024-02-24
Send SIGHUP to `yubikey-agent` if it's running and the connection fails

Ma27:
`yubikey-agent` is an SSH agent written in Go that uses the PIV module of a YubiKey for SSH[1].

Since it takes a persistent transaction on the YubiKey, using e.g. `ykman` will fail when it's active. I think it's tedious to always send a SIGHUP to it whenever I want to use `ykman`, so I added this small patch that does that for me. Inspired by what `age-plugin-yubikey` is also doing[2].

Not sure about Windows support, but I left it out because `yubikey-agent` doesn't do any Windows AFAIK and SIGHUP isn't supported there.

[1] https://github.com/FiloSottile/yubikey-agent
[2] https://github.com/str4d/age-plugin-yubikey/commit/1913838f8ed4b30c756c3c20ea5fdf1680ea97ca

### DIFF
--- a/ykman/pcsc/__init__.py
+++ b/ykman/pcsc/__init__.py
@@ -41,6 +41,10 @@ from time import sleep
 import subprocess  # nosec
 import logging
 
+import os
+import psutil
+import signal
+
 logger = logging.getLogger(__name__)
 
 
@@ -95,7 +99,7 @@ class ScardYubiKeyDevice(YkmanDevice):
         try:
             return ScardSmartCardConnection(self.reader.createConnection())
         except CardConnectionException as e:
-            if kill_scdaemon():
+            if kill_scdaemon() or kill_yubikey_agent():
                 return ScardSmartCardConnection(self.reader.createConnection())
             raise e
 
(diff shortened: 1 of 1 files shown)


dainnilsson:
Thanks! We do already have something similar in place for scdaemon (part of gpg) and I would prefer if this addition was done in the same way and place, without introducing new project dependencies. See https://github.com/Yubico/yubikey-manager/blob/main/ykman/pcsc/__init__.py#L98 for details.

Ma27:
@dainnilsson fair point! Updated accordingly! :)

dainnilsson:
Thanks! It's in the right location now, but I still think we should be able to do this without adding the psutil dependency.

Ma27:
Fair. Will look into it soonish and report back :)

Ma27:
Right, we depend on procps being installed here, so it's effectively just `pkill -HUP`. Updated the PR @dainnilsson .

dainnilsson:
Thanks. Th ecode looks fine, but you still have some unused imports that need to be removed.

PR #605, opened 2024-03-12
christopher/add fascn

christopherL91:
- **Update documentation**
- **Update typings**
- **Add dependency bitarray for FASC-N**
- **Add support for generating FASC-N**
- **Add tests for FASC-N**


christopherL91:
I found some typings errors in the code. I could move those commits out of this PR if you'd like.

christopherL91:
The module name is not set in stone, as there's already a piv.py it felt weird to have a module called the same. We could also discuss appropriate error handling and assertion that the lenght of each field are correct.

chrismeyer-yubico:
just checking if there is any progress on this PR?

dainnilsson:
No updates.

dainnilsson:
Replaced by #628

PR #709, opened 2026-01-16
Add the CXF decompress method for X509 certificates

manuco:
Our key management system uses certificates in a fairly specific format corresponding to the "CXF" format defined in the draft [The Compressed X.509 Certificate Format](https://datatracker.ietf.org/doc/html/draft-pritikin-comp-x509-00).

Although this draft has not been elevated to RFC status and is now expired, the described format has been implemented by some current systems.

The purpose of this PR is to provide a simple way for Yubikey holders to read this certificate format using standard tools. It does not replace the historical operation of certificates decompression but adds two additional decompression methods selected via a simple heuristic: the first method that works provides the decompressed certificate.

### DIFF
--- a/yubikit/piv.py
+++ b/yubikit/piv.py
@@ -420,7 +420,7 @@ _FASCN_LENS = (4, 4, 6, 1, 1, 10, 1, 4, 1)
 
 
 _cxf_dictionary = bytes([
-    # This data embed pre-computed structures used by the zlib deflate algorithm
+    # This data embeds pre-computed structures used by the zlib deflate algorithm
     # It's purpose is to avoid to store the full dictionary in the compressed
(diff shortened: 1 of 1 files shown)


dainnilsson:
@manuco - We released this in 5.9.0, but some more investigation seems to show that Net iD may not be using this scheme after all. We found that the sample we have has `FDICT=0` in the header, which means the decompress call ignores the `zdict` parameter. Removing the custom dictionary completely still results in successful decompression of the certificate.

Do you have any reason to believe that CXF is actually used by any real world application? I haven't seen any documentation describing the Net iD format. As it stands, we will likely drop this in the next release and just support what seems to be a proprietary Net iD format (01 00 L1 L2 <ZLIB data>) instead.

